PR #4, opened 2015-04-25
Make fonts compatible with MIT license

duncankl:
This PR removes the two WOFF fonts that would have required exceptions if we wanted to use an MIT license.

- Concourse is replaced with Fira Sans as my license wouldn't have allowed others to use it for their own projects.
- FF Tisa is replaced with ITC Charter to match the new stickers. Rather than including a WOFF, the OONI wordmark is rendered as SVG paths. I'm assuming your license allows it to be embedded as a "read only" subset but not distributed as the full font, but that would be good to double check.


hellais:
Sounds good. Merging.

PR #7, opened 2015-05-04
Fix empty page titles

duncankl:
Adding this dot allows the `head` partial to access the titles of blog posts and the home page.


hellais:
Ah awesome!

Was wondering what was going on there.

PR #8, opened 2015-05-04
Host the "fork on GitHub" image locally

duncankl:
To avoid a dependency on GitHub's image proxy which may change its URL scheme in the future.

Aside: This ribbon makes some of the nav links inaccessible on mobile. Would it be enough to make the "Code" link point here instead and maybe put "GitHub" in the title to make it obvious?


hellais:
Hum yeah for the moment I will merge this, though perhaps we should consider removing the ribbon or making it something that doesn't screw up the design on mobile.

I don't think the code should point to github, since the canonical code repo should be the one on torproject.org

PR #14, opened 2015-07-24
Translation in italian of adina announce

bassosimone:
Initial work by me, follow up work by Nexa colleagues.


hellais:
Excellent!

bassosimone:
As a reminder, **let's not delete the branch** (at least for a grace period) because it was tweeted

PR #15, opened 2015-07-24
Update ADINA event page

anadahz:
* Correct typos, syntax and format errors
* Modify text alignment to left (look nicer and easier to read)
* Update contact info

### DIFF
--- a/content/event/adina15.md
+++ b/content/event/adina15.md
@@ -68,10 +68,11 @@ Got an idea? Want to build it? We want to hack with you!
 
 Making internet censorship data (and, more generally, network measurements)
 easier to understand and study is crucial to foster an informed democratic
-debate on the governance of the internet. Democratic societies sometimes need
-to restrict access to certain content to safeguard the public interest; this
-leads to interesting research and policy questions such as how to transparently
-implement such means.
+debate on the governance of the internet. In fact, measuring content blocking
+is functional to quantify its collateral damage on citizens' digital rights and
+this information ultimately helps democratic societies to find the right
+balance between digital sovereignty, on the one hand, and fundamental rights,
+on the other end.
 
 ---
 
--- a/content/event/adina15.md
+++ b/content/event/adina15.md
@@ -71,8 +71,7 @@ easier to understand and study is crucial to foster an informed democratic
 debate on the governance of the internet. Democratic societies sometimes need
 to restrict access to certain content to safeguard the public interest; this
 leads to interesting research and policy questions such as how to transparently
-implement such means and how to openly verify that only what should be blocked
-is blocked.
+implement such means.
 
 ---
 
@@ -92,10 +91,9 @@ coding tasks.
 The hackathon will run from October 1st to October 2nd 2015.
 
 
-### October 1st
-Day 1
+### October 1st (Day 1)
 
-09:00 Start of event (Organizers should be there at 8 AM)
+09:00 Start of event (**Organizers** should be there at **08:00**)
 
 10:00 Introduction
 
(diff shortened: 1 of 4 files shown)


content/event/adina15.md
```diff
@@ -71,8 +71,7 @@ easier to understand and study is crucial to foster an informed democratic
 debate on the governance of the internet. Democratic societies sometimes need
 to restrict access to certain content to safeguard the public interest; this
 leads to interesting research and policy questions such as how to transparently
-implement such means and how to openly verify that only what should be blocked
-is blocked.
+implement such means.
```
bassosimone: I think we're losing a bit of meaning here that we should not loose. What about saying instead "transparently implement (and openly verify) such means"?

themes/ooni/static/css/event.css
```diff
@@ -54,7 +54,7 @@ nav.event-navbar {
 }
 
 .event-content {
-  text-align: center;
+  text-align: left;
```
hellais: I am not sure about changing this alignment. The designers have so far not gotten back to me on the current version of the site, but I would prefer to have their feedback before doing changes to the layout of the page.

hellais: Moreover doing left align makes the banner become off centered so that would also need to be changed:

screen shot 2015-07-24 at 3 18 20 pm

Also the join button perhaps needs to be adjusted.

content/event/adina15.md
bassosimone: or event better "... such content-blocking means"

anadahz: I think both proposed terms add over-complication and are hard to understand.

themes/ooni/static/css/event.css
anadahz: I can't see how the alignment of the banner is affected in your screenshot and in my screen.

content/event/adina15.md
bassosimone: I agree that it's not worded in a simple way. But it's quite important to keep the information that a number of research questions can arise from the data. And that's why I am insisting.

What about: "questions. Two possibile research questions are, for example, how to transparently implement such content-blocking means and how to verify that no content is blocked by mistake"?

themes/ooni/static/css/event.css
bassosimone: I surely don't qualify as an expert graphic, but I also prefer left alignment over center alignment

bassosimone:
I'm going to merge this in a few minutes, mainly to pull the improved wording, and since there is rough consensus towards left alignment.

content/event/adina15.md
bassosimone: After a follow up discussion, @anadahz and I agreed on a rewritten sentence that has now been committed on top of this pull request.

hellais:
I perfectly agree with the content changes, but strongly disaggree with the allignment change, at least in this current form.

By placing the text left alligned it is clearly misalligned wrt the logos in the header.

Could you please revert the left align?

On 26 July 2015 16:35:11 CEST, Simone Basso <notifications@github.com> wrote:
>I'm going to merge this in a few minutes, mainly to pull the improved
>wording, and since there is rough consensus towards left alignment.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/TheTorProject/ooni-web/pull/15#issuecomment-124997746

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

bassosimone:
Sure, see #16.

hellais:
To be clear, I am not saying we should not change the alignment, but these changes must be done in a separate pull request and not joined together with content changes.

I think perhaps a good compromise could be to leave the headers center aligned and make the text left aligned.
That will however still require re-aligning the banner too.

hellais:
Here is a picture to illustrate what I mean:

50affafc-3217-11e5-847c-e829c3eb065f

anadahz:
I needed this red aid lines to understand what you mean!
TBH the logo and picture are OK together but the text is awfully alligned in such an extent that it's almost unreadable!
Have you tried to read the text once for the page?

PR #22, opened 2015-08-31
Allow logging with Tor browser and strict settings

bassosimone:
Under certain strict settings Tor browser does not like it when you use sessionStorage and/or localStorage.

This diff removes any reference to such storages and uses instead a dictionary to keep authentication information.

With this diff committed on my staging system I was able to register myself for ADINA'15 from Tor browser.

This was joint work with @alemela and @davideallavena.

Any thoughts?

### DIFF
--- a/themes/ooni/static/js/adina15.js
+++ b/themes/ooni/static/js/adina15.js
@@ -1,14 +1,39 @@
 'use strict';
+var urlBase = "https://ooni-adina15.herokuapp.com/api";
 
 angular
   .module('adina15', ["lbServices"])
+
   .config(['$httpProvider', function($httpProvider) {
         $httpProvider.defaults.useXDomain = true;
         delete $httpProvider.defaults.headers.common['X-Requested-With'];
     }
   ])
   .config(["LoopBackResourceProvider", function (LoopBackResourceProvider) {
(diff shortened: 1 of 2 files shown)


0xPoly:
@bassosimone on Tor Browser 5.0.2, with the following privacy settings:

> [ ] always use private browsing
> [ ] remember history
> [ ] remember search/form history
> [x] accept cookies from sites
> accept third party: never
> keep until: they expire

the patch does *not* work. I still get an error message saying 'authorization required'.

bassosimone:
@0xPoly can you try again now? There was one additional issue that TBB removes the 'Authentication' header in some cases.

With this diff in, I was able to register using TBB and a configuration similar to yours.

0xPoly:
It works now. Thanks for the quick reply!

hellais:
This is excellent. Thanks a lot. Will review and deploy this most probably tomorrow.

bassosimone:
Awesome!

bassosimone:
We've improved the diff a bit to reduce the amount of code in `lb-services.js`. Still no idea regarding how to avoid the `localStorage` and `sessionStorage`.

hellais:
This is good.

hellais:
Apologies for not reviewing it sooner. Thanks @DavideAllavena @bassosimone @0xPoly for the great work on this. Few lines, much work.

PR #27, opened 2015-09-24
Added required fields: birthCity, birthDate, phoneNumber.

bassosimone:

PR #28, opened 2015-09-24
Add required fields, the hackish version

bassosimone:
This is same as #27 except that the code is structured to hackishly avoid to modify the heroku backend, storing all required fields in the twitter field.

The difference between this and #27 is bd201dbc529 (i.e. the hackish commit).


bassosimone:
Unfortunately this is not doable, because of CORS. The heroku server uses `https://ooni.torproject.org/` as origin and we cannot undo that.

PR #29, opened 2015-09-25
Use the -staging heroku database

bassosimone:
We are not able to access the non staging database currently, luckily the -staging one is quite in sync and already contains the changes required to contain the newly requested fields.


anadahz:
Confirmed switching to the staging database.

PR #31, opened 2015-10-01
Fix/adina15 event

anadahz:


bassosimone:
Looks good to me, thank you @anadahz!

PR #32, opened 2015-10-10
Fix baseURL config option

anadahz:


bassosimone:
let's see what is the effect of this :)

PR #33, opened 2015-10-28
Remove use of .Site.BaseURL, replace with relative URLs

vtduncan:
Currently the site has some HTTPS mixed-content warnings. Looks like it was caused by a build with BaseURL set to `http://ooni.torproject.org/` rather than `https://ooni.torproject.org/`.

This commit removes the use of BaseURL and replaces it with relative URLs, so the site continues to work at both:

- https://ooni.torproject.org/ and
- https://thetorproject.github.io/ooni-web/

Note that there are still absolute URLs used for images etc. in blog posts. This breaks them on https://thetorproject.github.io/ooni-web/. However replacing these with relative URLs would break them in the RSS feed instead. For the time being I have left them as is.

### DIFF
--- a/themes/ooni/layouts/_default/single.html
+++ b/themes/ooni/layouts/_default/single.html
@@ -5,7 +5,7 @@
     <nav>
       <div class="col-1">
         <a href="/">
-          <img src="{{ .Site.BaseURL }}images/ooni-header-mascot.svg" width="25" height="25"/>
+          <img src="../../images/ooni-header-mascot.svg" width="25" height="25"/>
           <h1>OONI</h1>
         </a>
       </div>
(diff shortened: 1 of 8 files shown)


bassosimone:
Thanks, @vtduncan! I will review and test this soon and report back!

hellais:
LGTM :+1:

hellais:
(it's a bit of a hack we should actually use the new syntax, but it's good enough :smile: )

hellais:
Restored ooni.io :)

hellais:
It's alive: https://ooni.torproject.org/!

vtduncan:
Yay! Glad that worked.

PR #37, opened 2016-03-07
Convert SVG images to PNG on the homepage and in headers

vtduncan:
This fix is for Tor Browser. It doesn't seem to display SVGs on the homepage anymore, leaving some rather noticeable blank rectangles :)


hellais:
Sounds good, thanks :)

PR #38, opened 2016-03-16
Fix text alignment in Lepidopter template

vtduncan:


hellais:
This should be rebased onto master and probably use the png versions instead of the SVG versions of the image due to #37

vtduncan:
Ok, this should be ready to go now. It is rebased and includes the SVG → PNG change.

hellais:
LGTM!

PR #42, opened 2016-05-04
Add contact details

hellais:
This adds the contact details to the homepage.


bassosimone:
The meaning of life is here, visible to everyone. Except that it's encrypted.

PR #47, opened 2016-05-08
Remove ADINA leftovers from the website

anadahz:
The ADINA pages are still available online and should be removed.
We can enable the event page(s) when we have a new event.


hellais:
LGTM.

PR #53, opened 2016-05-23
Add image links, verification doc, fixed names, format

anadahz:
* Add lepidopter download link
* Add image verification docomentation
* Replace generic image names with the actual name of the image
* Format fixes

### DIFF
--- a/content/lepidopter/lepidopter.md
+++ b/content/lepidopter/lepidopter.md
@@ -3,7 +3,7 @@ date: "2015-09-15"
 title: "Lepidopter Installation: Help Guides and Resources"
 description: "Helpful resources and guides for getting started with Lepidopter
 Raspberry Pi image"
-tags: [ "lepidopter", "image", "raspberry pi" ]
+tags: [ "lepidopter", "image", "raspberry pi", "distribution" ]
 ---
 
 # Lepidopter quick start guide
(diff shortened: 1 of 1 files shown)


content/lepidopter/lepidopter.md
```diff
@@ -18,11 +18,56 @@ Lepidopter in QEMU is described
 
 ## Download Lepidopter Raspberry Pi image
 Lepidopter Raspberry Pi image is compatible with all types of Raspberry Pi
-(tested: B+, RPi 1 model B, RPi 2 model B). [Lepidopter image download](XXX)
+(tested: B+, RPi 1 model B, RPi 2 model B). [Lepidopter image download]
+(https://measurements.ooni.torproject.org/lepidopter/lepidopter-alpha-armel.img.xz)
```
hellais: I think we should have another host that is not `measurements.ooni` for hosting raspberry pi and other distributions of ooniprobe.
Maybe something like: dist.ooni.torproject.org?

content/lepidopter/lepidopter.md
```diff
@@ -77,7 +122,7 @@ delete your primary Linux partition. Please be careful.
    should first [Download Lepidopter Raspberry Pi image]
 (#download-lepidopter-raspberry-pi-image:7565a061641bd1e599d91e306fc304d1)
 1. Extract the image, with:
-   ```xz --decompress ~/lepidopter.zip```
+   `xz --decompress --verbose --no-sparse lepidopter-alpha-armel.img.xz`
```
hellais: Can we make the image in gzip format instead of `xz`. I believe xz doesn't come by default with most distributions.

content/lepidopter/lepidopter.md
anadahz: True,  I just wanted to publish the image and that was the fastest way.

Perhaps it makes sense to have a downloads.ooni.torproject.org or something similar where all OONI related programs and tools can be downloaded?

content/lepidopter/lepidopter.md
anadahz: Yes it makes sense but if we go this way we should perhaps use `zip` as compression format and maybe add both `zip` and `xz` formats?

PR #59, opened 2016-06-17
Add ISP AS numbers, update raw measurement URLs

anadahz:


anadahz:
I'm going to proceed and merg this PR.

PR #60, opened 2016-06-24
Add contribute/help section, default credentials, fix format/typos

anadahz:

PR #68, opened 2016-09-08
Doc/lepidopter services

anadahz:
This PR is based in #60 that contains important documentation and fixes.


content/install/lepidopter.md
```diff
@@ -465,8 +468,82 @@ yes and press enter.
...
+the Bonjour software.
+
+### Windows
+
+By default Windows systems have no built-in support for mDNS/DNS-SD. However
+some third-party applications come bundled with the Bonjour software, service
+that supports automatic discovery of hosts in the local network.
+
+The applications that are bundled with the Bonjour software in Windows:
+
```
hellais: Include a download link to install the software and steps on how a windows user can then use bonjour once it's installed.

content/install/lepidopter.md
```diff
@@ -465,8 +468,82 @@ yes and press enter.
...
+## Service discovery install instructions
+
+### Linux
+
+By default most Linux distributions (Debian, Fedora, Ubuntu, Archlinux) use the
+Avahi daemon by default. In case that your system doesn't come with the Avahi
+daemon pre-installed you should be able to install it via the package
+management of your distribution, the name of the package is usually something
+similar to `avahi-daemon`.
```
hellais: Add instructions on how to check if avahi-daemon is installed, what to do if it's not installed and how to check if it's installed how to ensure it's working properly.

content/install/lepidopter.md
```diff
@@ -465,8 +468,82 @@ yes and press enter.
...
+
+## Access ooniprobe's web interface
+
+If you have the required dependencies you should be able to access
+ooniprobe web interface by entering the following URL to your web browser:
+(http://lepidopter.local:8842)
+
+If you can't access ooniprobe's [web interface URL]
+(http://lepidopter.local:8842) please follow the installation [instructions]
+(#service-discovery-install-instructions) related to your operating system.
```
hellais: Maybe mention also as fallback the http://lepidopter:8842/ address in case the local resolve automatically sets up records.

content/install/lepidopter.md
```diff
@@ -465,8 +468,82 @@ yes and press enter.
...
+
+Lepidopter uses a mDNS (multicast DNS Service Discovery) (Avahi) to publish the
+SSH and ooniprobe web interface services on a local network with no extra
+configuration or the need to find out the IP address of lepidopter host.
+
+## Access ooniprobe's web interface
+
+If you have the required dependencies you should be able to access
+ooniprobe web interface by entering the following URL to your web browser:
+(http://lepidopter.local:8842)
```
hellais: This should be a link

content/install/lepidopter.md
anadahz: Do we really need to add links of these application that most probably change?

content/install/lepidopter.md
anadahz: I guess this is out of scope, however all mentioned recent OS releases include the avahi daemon service.

content/install/lepidopter.md
anadahz: Under which circumstances is this happening?

content/install/lepidopter.md
anadahz: Thanks changed to link.

content/install/lepidopter.md
hellais: I think we should say:

If you have any of the following applications installed, then you don't need to do anything. However if you don't then install XXX, where XXX probably should be the Bonjour Print services.

If the links to them change in the future we should update them, but it's important that we give some instructions that a newbie can follow to access their pi.

content/install/lepidopter.md
hellais: I don't think it's out of scope. If we are providing instructions on how to set it up or install it, we should mention how a linux user does this in the form of a tutorial.

content/install/lepidopter.md
```diff
@@ -465,8 +468,82 @@ yes and press enter.
...
+If you can't access ooniprobe's [web interface URL]
+(http://lepidopter.local:8842) please follow the installation [instructions]
+(#service-discovery-install-instructions) related to your operating system.
+
+## Access lepidopter SSH service
+
+You should be able to access lepidopter's SSH service under the following
+hostname: `lepidopter.local`
+
+## Service discovery install instructions
```
hellais: The order of these should probably be inverted. 1. Windows 2. Mac OS X 3. Linux.

content/install/lepidopter.md
hellais: When the resolver of the router sets up a record for the hostname "lepidopter".

content/install/lepidopter.md
anadahz: I follow the same order in the guide: https://github.com/TheTorProject/ooni-web/blob/master/content/install/lepidopter.md#setting-up-your-raspberry-pi-with-lepidopter
ie. 1. Linux, 2. Mac OS X, 3. Windows

anadahz:
I have updated the docs based on feedback from @hellais .

anadahz:
@hellais I made the changes in  aaabcee for https://github.com/TheTorProject/ooni-web/commit/012ed0be5c2554082d3db365793bff5990d97b9c and also updated the documentation to the reflect the changes of newer lepidopter image build.

content/install/lepidopter.md
```diff
@@ -110,18 +110,18 @@ uid   Lepidopter Team (signing key)
...
 
 You should look for the message `Good signature from "Lepidopter Team (signing
 key)"`:
 
 ```
-gpg: assuming signed data in `lepidopter-alpha-armel.img.zip'
+gpg: assuming signed data in `lepidopter-v0.3.5-beta-armel.img.zip'
 gpg: Signature made Sun 22 May 2016
 gpg:                using RSA key 0xBA56AC5A53E9C7A4
 gpg: using PGP trust model
```
darkk: - I see trailing newline in the paste after `gpg: binary signature, digest algorithm SHA512`
- what about using full fingerprint like `--recv-keys 625511968E240F24F6CFD0B6BA56AC5A53E9C7A4` ?

content/install/lepidopter.md
```diff
@@ -266,7 +266,7 @@ the name of the whole SD card as described above, not just a partition of it
 (for example, sdd, not sdds1 or sddp1, or mmcblk0 not mmcblk0p1).
 
 ```     
-dd bs=4M if=~/lepidopter-alpha-armel.img of=/dev/sdd
+dd bs=4M if=~/lepidopter-v0.3.5-beta-armel.img of=/dev/sdd
```
darkk: Don't one need `sudo` to run `dd` against SD card?

content/install/lepidopter.md
```diff
@@ -216,16 +216,16 @@ machine. If you specify the wrong device in the instructions below you could
...
 
-   `unzip lepidopter-alpha-armel.img.zip`
+   `unzip lepidopter-v0.3.5-beta-armel.img.zip`
 
 If you have downloaded the [xz compressed image]
-(https://get.ooni.torproject.org/lepidopter/lepidopter-alpha-armel.img.xz):
+(https://get.ooni.torproject.org/lepidopter/lepidopter-v0.3.5-beta-armel.img.xz):
 
-   `xz --decompress --verbose --no-sparse lepidopter-alpha-armel.img.zip`
+   `xz --decompress --verbose --no-sparse lepidopter-v0.3.5-beta-armel.img.zip`
```
darkk: - `xz` can't decompress `zip`
- Why `--no-sparse` ? Saving 3Gb of zeros looks like a good idea to me.

content/install/lepidopter.md
```diff
@@ -74,13 +74,13 @@ If your laptop does not have an SD card slot, please ensure that you have an SD
...
-(https://get.ooni.torproject.org/lepidopter/lepidopter-alpha-armel.img.zip)
+(https://get.ooni.torproject.org/lepidopter/lepidopter-v0.3.5-beta-armel.img.zip)
 
 Consider downloading the [xz compressed image]
-(https://get.ooni.torproject.org/lepidopter/lepidopter-alpha-armel.img.xz)
-for a significantly reduced file size (ca. 114M smaller) compared to the zip
-archive. Note that this will require an extra program to be installed depending
-on your OS.
+(https://get.ooni.torproject.org/lepidopter/lepidopter-v0.3.5-beta-armel.img.xz)
+for a significantly reduced file size compared to the zip archive.
```
darkk: IMHO exact sizes are better.

* Lepidopter [image.zip, 254 MB](https://get.ooni.torproject.org/lepidopter/lepidopter-v0.3.5-beta-armel.img.zip), [GPG signature](https://get.ooni.torproject.org/lepidopter/lepidopter-v0.3.5-beta-armel.img.zip.asc)
* Lepidopter [image.xz, 174 MB](https://get.ooni.torproject.org/lepidopter/lepidopter-v0.3.5-beta-armel.img.xz), [GPG signature](https://get.ooni.torproject.org/lepidopter/lepidopter-v0.3.5-beta-armel.img.xz.asc)

content/install/lepidopter.md
```diff
@@ -74,13 +74,13 @@ If your laptop does not have an SD card slot, please ensure that you have an SD
...
+Note that this will require an extra program to be installed depending on your
+OS.
```
anadahz: You are right it makes sense to add this.

content/install/lepidopter.md
anadahz: I recall some issues with different version of dd not parsing correctly sparse files.

content/install/lepidopter.md
anadahz: I was hoping to have a link pointing to the latest image file and having inaccurate exact file sizes numbers is not nice.

content/install/lepidopter.md
anadahz: Ack!

anadahz:
@darkk thanks for your comments, add fixes in 317a8bb.

hellais:
So I have thought a bit about this and I have the following feedback.

I think the installation instructions for lepidopter should be restructured to take into account the fact that there are two types of users of lepidopter (or maybe even 3):

1. A user of lepidopter that has gotten it via our partnership program. These from what I can tell are actually the majority of lepidopter users and are people that don't care how to burn the image to an SD card. This user is not technically savvy.

2. A user of lepidopter that wants to download the image and burn it onto their SD card. These are a minority of users that are quite technically savvy.

3. (if we include this as well) Power users or developers that want to contribute in some way to the project. Very technically savvy.

In light of this I think the guide should be divided into at least 2 sections.

* User guide

* How to burn an image

Inside the user guide there should be clear and simple explanations as to how to connect the raspberry pi to the power and internet (maybe even including screenshot). @andresazp showed me the document he created for VI and it looks great.
It should also cover information on how to initialise the pi (answering the informed consent) and check that measurements are being run, by visiting the web UI.

An outline of the User Guide could be:

* Setting up the device (covering how to connect it to internet and power, putting in the SD card etc.)

* First start (covering how to reach the pi over the network, using lepidopter or lepidopter.local) and answering the informed consent

* Checking that everything works (checking that measurements are being run)

The How to burn an image can remain as the current section titled: "Downloading and verifying the Lepidopter Raspberry Pi image"

I would also try to make this guide overall much shorted and where possible point to external resources instead of having so much content in here (for example explaining the tests).

hellais:
I am also thinking that maybe the whole part on burning the image could be tucked inside of another page entirely as it makes the whole guide much less user friendly.

Another comment is that the whole section "Lepidopter services" is far too technical and contains a bunch of information that will not be useful to most cases (why would the normal user care to know that there is a mDNS service for SSH?).

I would tuck this stuff as well inside of some advanced section and try to have a minimal guide that is easy for most users to follow.

The experience we had with real deployment of these is that even just explaining how to plug in a pi into a router and connecting to power can be challenging for some non technically savvy users and since this guide is addressed to them we should take that into account.

anadahz:
@hellais your comments:https://github.com/TheTorProject/ooni-web/pull/68#issuecomment-256102151 and https://github.com/TheTorProject/ooni-web/pull/68#issuecomment-256105710  make sense but they will be better addressed in another PR as we need to make the release announcement of ooniprobe v2.0.0 and WUI (https://github.com/TheTorProject/ooni-web/pull/76) ASAP since the users are still using older versions of lepidopter and ooniprobe.

hellais:
Filed an issue with my feedback.

Merging this into the website.

PR #69, opened 2016-09-16
Doc/release announcement

anadahz:
Do you think that we should add more screenshots, to guide the user through the
setup wizard?


hellais:
This was written with the goal of being just the pre-release announcement that we do only to the developer mailing lists and not meant to be a blog post itself.

I believe @agrabeli is writing a full length blogpost that will come out after 1 week of the release candidate initial testing phase to announce the official release of ooniprobe 2.0.0.

I would say we use this text as content for mailing lists, do a round of testing and then do a bigger announcement for the actual release.

hellais:
I am closing this since we used this text in mailing lists.

PR #89, opened 2016-11-12
Fix/homepage

p1g30n:
Various Fixes to the Homepage, especially in mobile view


themes/ooni/static/css/master.css
```diff
@@ -265,10 +267,9 @@ th, td {
   font-weight: inherit;
 }
 
-pre {
```
hellais: I suspect this could create issues in other places where we are using "pre".

PR #91, opened 2016-11-13
add faq

juga0:
The CSS width for the element nav>div.col-2 must be increased because with the new link there is not space and the navigation bar get split in 2 rows.


hellais:
It seems like the menu bar "FAQ" is not being marked active when clicked on it.

agrabeli:
This FAQ is now out of date. I'll be re-writing this to also include a lot of other FAQ.

agrabeli:
I'm closing this PR because this version of the FAQ is outdated and it's going to be replaced by the new FAQ: https://github.com/TheTorProject/ooni-web/pull/301

PR #94, opened 2016-11-15
Update ooniprobe installation guide

anadahz:
* Update pip and vagrant install methods
* Add torproject repository install instructions for Debian/Ubuntu
* Update running ooniprobe section


anadahz:
Dependency: ooniprobe package should be in `deb.torproject.org`.

content/install/ooniprobe.md
```diff
@@ -44,45 +44,53 @@ as your employer, ISP or government). Please read our relevant
...
+
+**On Debian stable (jessie):**
+
+**Step 1.** Configure the torproject repository by typing the following in your terminal:
 
 ```
-(crontab -l 2>/dev/null; echo "@daily ooniprobe $THE_OONI_COMMAND") | crontab -
+gpg --keyserver keys.gnupg.net --recv A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
+gpg --export A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89 | sudo apt-key add -
+echo 'deb http://deb.torproject.org/torproject.org jessie main' | sudo tee -a /etc/apt/sources.list
```
hellais: Can we put these inside of `/etc/apt/sources.list.d/ooniprobe.list` so that we can make it easier to provide updates to the install instructions in the future?

I think it's easier to delete and rewrite a file, than it is to grep for the line containing the ooniprobe source list inside of `sources.list`

content/install/ooniprobe.md
```diff
@@ -44,45 +44,53 @@ as your employer, ISP or government). Please read our relevant
...
 
+**Step 2.** Type the following in your terminal:
 
-### Linux: Debian
+```
+sudo apt-get install ooniprobe deb.torproject.org-keyring
+```
 
-**On Debian testing and unstable:**
+**On Debian (contrib) testing and unstable:**
```
hellais: We should advise people on testing and unstable to also use the torproject debian mirror. That way there is no confusion about contrib or non-contrib.

content/install/ooniprobe.md
```diff
@@ -335,26 +241,25 @@ sudo apt-get install git
...
-git clone https://github.com/TheTorProject/ooni-backend.git
 cd ooni-probe/
 vagrant up
 ```
 
 **Step 5.** Access your virtual machine with:
 
 ```
-vagrant ssh probe
+vagrant ssh
```
hellais: When using vagrant you still need to specify probe after vagrant ssh since it's a multimachine VM

content/install/ooniprobe.md
anadahz: To avoid further confusion it makes more sense to remove testing and unstable since there are no "standard" Debian distributions.

anadahz:
Thanks for your comment @hellais I have updated the PR.

hellais:
> Remove testing and unstable install instructions to avoid confusion about contrib and non-contrib repositories

I think we should leave also installation instructions for testing and unstable (since we support those too), but perhaps put them after debian jessie.

There are people that run debian testing and they may be interested in installing it as well.

anadahz:
@hellais added install instructions for testing and unstable in 87cbbfb.

irl:
I would really prefer it if users were installing from the Debian archives on Debian. It means 5 days between the upload to unstable and the upload to stable (assuming all dependencies are already backported) but this allows users to not have to enable a second repository and also doesn't introduce another thing they have to bootstrap trust for.

The reason I wanted to use deb.torproject.org was that it's nicer than a PPA and we can use existing tpo infrastructure, but I never intended that end users using Debian would be using it. The option of having speedier access to stable updates is there on deb.tpo but this could lead to breakage in stable systems, as the 5 days of QA testing hasn't happened.

anadahz:
@irl I agree that's way better to have stable packages in Debian, however users are going to install the tor package from the deb.torproject.org repository, and by default (without any pinning) the default installation candidate package of  ooniprobe is going to be installed from the deb.torproject.org repository.  

Perhaps we should use the same model for testing and distributing (stable) ooniprobe packages in deb.torproject.org similar to tor, tor-nightly and tor-experimental packages and move ooniprobe packages to the "stable" releases given enough time for the QA testing?

irl:
@anadahz This is what I discussed with weasel, and he has pointed me at the things I need to do to make this happen. Work in progress.

hellais:
@irl @anadahz I don't fully understand the implications of what you are discussing here, though what I would like to happen is that once the user has done the first install of ooniprobe, whenever we cut a new release we should be able to push an update to them with a version we consider good for them.

Is this going to be possible with the approach you are discussing here?

Ideally it would also be the case that they don't have to mess to much up with their repository settings, so adding a single repository is easier (in my opinion) than having them adjust to use backports and/or pin certain packages on another release of the repo.

irl:
Things get built daily, and go to the daily branch, to catch problems with test failures in the different suites primarily. When you tag a release it goes into a proposed-branch and then when tested it can be promoted to the user-facing suite. Nothing ships automatically.

For Debian, those packages would then be rebuilt for Debian, which is the reverse of what currently happens where the deb.tpo packages are based on the Debian package.

If users are running Debian, I would prefer it if we ask them to enable backports to ensure that all the backports required are tested well. Currently, I'm backporting selected packages that are required by ooniprobe but in some cases crippling the build of the package. An example is python-ipaddress that in Debian jessie-backports has python and pypy builds, but on deb.tpo only has the python build. This shouldn't be a problem but it's not tested extensively.

The backports problem is unavoidable for Ubuntu, but is avoidable for Debian as those packages have already been backported, either by me or by their original maintainer, and their dependencies backported too so that the packages don't have to be crippled in order to kill extraneous dependencies.

hellais:
@irl I understand your concerns regarding how well tested the software is, but maybe we can internally agree on some policy whereby we don't push out a package to the deb.torproject.org repository until it has gone through the debian testing system?

That said I would be much happier if we had people just do one thing when they need to install ooniprobe and it not changing over time.

For example currently ooniprobe is inside of contrib, so they would have to enable contrib if they are on testing and if they are running stable they would have to add backports and contrib.

I think that in order to keep the installation as simple as possible it would be best if we could keep the steps needed to install as simple and straightforward as possible.

Ideally I would want people to just have to do (in debian based systems):

1. Add deb.torproject.org repository to their /etc/sources.list/

2. apt-get install ooniprobe

Ideally there would not be any distinction between debian stable, testing, unstable and ubuntu, wily, xenial, etc. a part from changing the name of the release in the apt sources line.

Regarding backporting of dependencies, wouldn't this is something that would have to be done anyways? If so is there extra effort required to add an already backported package to the torproject deb repo?

hellais:
Merging!

PR #98, opened 2016-11-21
ooni-web /nettest/ fixes

p1g30n:
- Fixed /nettests/ boxes by wrapping the content elements into divs
- Edited css accordingly
- Removed duplicate slash in URL path


p1g30n:
just seeing that the media query at the bottom of master.css should be changed to '699px' instead of '48em' for consistency

hellais:
Great stuff. Thanks for this!

(I merged it into master, but for some reason github doesn't figure it out. Maybe it has something to do with the fact that there were some merge commits before this that I squashed when merging).

PR #99, opened 2016-11-25
Feature/re organise

hellais:


hellais:
I guess we can merge this.

PR #104, opened 2016-12-15
Add universe repository requirement to Ubuntu installations

anadahz:


hellais:
Can we use the apt-add-repository command instead? This way if it's already present we don't add duplicates.

anadahz:
The `apt-add-repository` command requires the `python-software-properties` package dependency which is also (as with the universe repository) [not installed by default](https://help.ubuntu.com/community/Repositories/CommandLine#Adding_Launchpad_PPA_Repositories) in all Ubuntu releases.

hellais:
From the page you linked:

> 3. Should be installed by default. On older or minimal Ubuntu releases, you may have to install software-properties-common and/or python-software-properties first (sudo apt-get install python-software-properties) 

Since we only support for ubuntu >= 14.04 I don't think it's a concern that it doesn't come by default on older releases. The fact that minimal installations don't contain I don't think is a big deal. Somebody running a minimal version of ubuntu is probably knowledgeable enough to add the sources themselves.

anadahz:
@hellais The package `python-software-properties` is _not_ installed by default in Ubuntu versions >= 14.04. From what I understand the Ubuntu releases that by default have the package `python-software-properties` installed, have the universe repository enabled by default.

Perhaps we can add a sub-section that the universe repository should be enabled for ooniprobe to be installed successfully. However this over complicates users that just copy/paste the install instructions.

hellais:
Hum ok. Here they claim it's part of the desktop version: https://bugs.launchpad.net/ubuntu/+source/ubuntu-meta/+bug/439566 and only recently (14.04 maybe?) part of server as well.

Anyways I just wanted to avoid people having duplicate lines in their apt source lists, but if you think this is not going to be robust enough we can proceed with this way.

PR #108, opened 2017-01-26
Updated Data Policy

agrabeli:
This pull request includes an update to OONI's Data Policy based on feedback and suggestions that we received through legal consultation. This updated version also takes into account various developments since the Data Policy was first drafted (May 2016), such as OONI's mobile app, measurements interface, web UI, etc.


anadahz:
Thanks, LGTM!

PR #122, opened 2017-03-17
Post/shutdowns

agrabeli:
This post provides an outline of OONI's basic methodology when examining internet shutdowns through public data sources.


darkk:
Seems, pictures are missing.

agrabeli:
@darkk They should appear when the post gets published since the images have already been published in previous OONI posts (and are included in this repo).

PR #129, opened 2017-03-30
Update key expiration time, change key ids from 32bit to 64bit

### LINKED ISSUE #128: PGP key appears to have expired on March 23rd, 2017
Page at https://github.com/TheTorProject/ooni-web/blob/master/themes/ooni/layouts/section/about.html (https://ooni.torproject.org/about/) says the PGP key expired on March 23rd, 2017. The same key is quoted in https://github.com/TheTorProject/ooni-web/blob/master/themes/ooni/layouts/index.html (https://ooni.torproject.org).

Also, I cannot seem to find the actual key anywhere but perhaps this is just me.

darkk:
Closes #128

### DIFF
--- a/content/about/data-policy.md
+++ b/content/about/data-policy.md
@@ -178,7 +178,7 @@ For questions and comments on this Data Policy, please contact:
 
 ```
 Key ID:
-    0CB177B7
+    6B2943F00CB177B7
 Fingerprint:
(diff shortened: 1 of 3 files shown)


anadahz:
LGTM

hellais:
It seems like the site was actually not updated.

If you merge stuff to master, but don't plan on deploying it be sure to mention it when you merge it so that somebody can do that...

anadahz:
@hellais ack!

PR #131, opened 2017-04-07
Add Telegram test documentation

anadahz:

PR #132, opened 2017-04-07
Update lepidopter installation, add tests, signature sections

anadahz:
* Add missing tests include in lepidopter version
* Update supported models
* Add signature and trust sections

### DIFF
--- a/content/install/lepidopter.md
+++ b/content/install/lepidopter.md
@@ -42,6 +42,30 @@ Designed to examine whether URLs are blocked or not (and if so, how)
 
 * **[HTTP header field manipulation](https://ooni.torproject.org/nettest/http-header-field-manipulation/):** Similar to the previous test (HTTP invalid request line)
 
+* **[Meek Fronted Requests]
+(https://ooni.torproject.org/nettest/meek-fronted-requests/):** Designed to
+examine whether the domains used by Meek (a type of Tor bridge) work in tested
+networks
+
+* **[Vanilla Tor](https://ooni.torproject.org/nettest/vanilla-tor/):** Designed
+to examine the reachability of the Tor network
+
+* **[Tor Bridges test]
+(https://ooni.torproject.org/nettest/tor-bridge-reachability/):** Designed to
+examine the reachability of Tor bridges in your network
+
+* **[WhatsApp test](https://ooni.torproject.org/nettest/whatsapp/):** Designed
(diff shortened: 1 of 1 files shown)


content/install/lepidopter.md
```diff
@@ -42,6 +42,30 @@ Designed to examine whether URLs are blocked or not (and if so, how)
...
+
+* **[WhatsApp test](https://ooni.torproject.org/nettest/whatsapp/):** Designed
+to examine the reachability of both WhatsApp and WhatsApp's web interface
+(web.whatsapp.com) in your network
+
+* **[Facebook Messenger test]
+(https://ooni.torproject.org/nettest/facebook-messenger/):** Designed to
+examine the reachability of Facebook Messenger in your network
+
+* **[Telegram test](https://ooni.torproject.org/nettest/telegram/):** Designed
```
anadahz: The URL on this depends on https://github.com/TheTorProject/ooni-web/pull/131

content/install/lepidopter.md
```diff
@@ -42,6 +42,30 @@ Designed to examine whether URLs are blocked or not (and if so, how)
...
+to examine the reachability of Telegram messenger and Telegram's web
+version in your network
```
hellais: This section actually applies to all of OONI and is not specific to lepidopter. We should probably remove the listing of the tests in this section and just point to: https://ooni.torproject.org/nettest/.

anadahz:
> I would also wait until 1.0.0 is released so we can put the most up to date version of lepidopter in here.
It's out: https://github.com/TheTorProject/lepidopter/releases/tag/v1.0.0

content/install/lepidopter.md
anadahz: > This section actually applies to all of OONI and is not specific to lepidopter.

I think is fine to mention what tests are being tested by each platform/program/application since not all tests can be run from all devices or software.

hellais: Though the tests on lepidopter are exactly the same as the desktop version of ooniprobe. To avoid having to update things in many places (i.e. when new tests are added to ooniprobe), we should just point to a common page.

anadahz: > To avoid having to update things in many places (i.e. when new tests are added to ooniprobe), we should just point to a common page.

Makes sense, let's do this when this page is created?
re: https://github.com/TheTorProject/ooni-web/issues/136

Issue #92, opened 2016-11-14
Add docker file to reproducibly build the site


hellais:
We should have a `Dockerfile` that allows us to build the website with a known good version of hugo and has everything setup properly to do it.

The running of the dockerfile should be instrumented from the `Makefile`.

darkk:
It's non-trivial to build same blob for https://ooni.github.io/ and https://ooni.torproject.org/ due to baseUrl being hardcoded into the webpage body. So two branches should be built: one for ooni.github.io and another one for ooni.tpo.

anadahz:
Partly implemented in: https://github.com/TheTorProject/ooni-web/pull/134

darkk:
I think, current Travis build documented in [tarvis.yml](https://github.com/TheTorProject/ooni-web/blob/master/.travis.yml) and [suggested workflow](https://github.com/TheTorProject/ooni-web/blob/master/.assets/ooni.io.png) are good enough to close this ticket.

PR #134, opened 2017-04-25
Add docker specific make targets to test ooni-web

anadahz:
This patch includes a Dockerfile and the relevant make targets to build
the ooni-web hugo website. One can test the ooni-web website by
visiting: http://localhost:1313/
Party implements:
https://github.com/TheTorProject/ooni-web/issues/92

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,10 @@ RUN wget https://github.com/spf13/hugo/releases/download/v${HUGO_VERSION}/hugo_$
   mv hugo /usr/bin/hugo
 
 # Clone required OONI repositories
-RUN git clone https://github.com/TheTorProject/ooni-web ${OONI_WEB_REPO_DIR} && \
-  git clone https://github.com/TheTorProject/ooni-probe ${OONI_PROBE_REPO_DIR}
+RUN git clone https://github.com/TheTorProject/ooni-probe ${OONI_PROBE_REPO_DIR}
 
 # Generate OONI website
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# Install hugo and package dependencies to build ooni-web
+RUN apk add --update wget ca-certificates make git py2-pip
+WORKDIR /tmp
+RUN wget https://github.com/spf13/hugo/releases/download/v${HUGO_VERSION}/hugo_${HUGO_VERSION}_linux-64bit.tgz && \
+  tar xzf hugo_${HUGO_VERSION}_linux-64bit.tgz && \
+  rm -r hugo_${HUGO_VERSION}_linux-64bit.tgz && \
+  mv hugo /usr/bin/hugo
+
+# Clone required OONI repositories
+RUN git clone https://github.com/TheTorProject/ooni-web ${OONI_WEB_REPO_DIR} && \
```
hellais: You should not be cloning the ooni-web repository, otherwise you will be getting a copy of master and not the branch you are currently on. This means one cannot use the docker container to test a feature they are currently working on, nor run the daemon to test live changes.

Instead it should be sharing the current working directory.

anadahz: Indeed, thanks for reviewing this, updated.

darkk:
#92 is closed now. Seems, this can be put on the shelf as well.

PR #135, opened 2017-05-03
Add F-Droid badge link to install section

anadahz:


hellais:
It looks like the 2x png is actually just the 1x sized one scaled up and as a result looks all grainy. Could you replace it with either an actually higher resolution one or just not include it at all?

hellais:
The upscaled image is now mis-aligned and the font of the "F-Droid" writing is not the same as the one in the original 1x image.

anadahz:
> The upscaled image is now mis-aligned and the font of the "F-Droid" writing is not the same as the one in the original 1x image.

This is actually based on the svg version in https://gitlab.com/fdroid/artwork/tree/master/badge

Perhaps it needs some "manual" intervention or @hellais how did you generated the rest of the badges?

hellais:
> Perhaps it needs some "manual" intervention or @hellais how did you generated the rest of the badges?

It's still broken. I will recreate the badged manually (the svg they uploaded is broken).

The other badges I just took the official ones that were well made and didn't require any changes.

PR #155, opened 2017-10-04
Updates to Catalonia blog

agrabeli:
Added the Catalan translation and an "updates" section (linking to Qurium's findings).


darkk:
Live.

PR #157, opened 2017-10-05
I18n

darkk:
It's mostly OK. I updated style to include cross-references to translations in **other** languages, it looks like that:
brazil-whatsapp 2017-10-05 14-48-11

English text references `pt` text. Old URL is preserved with `alias`.
brazil-whatsapp 2017-10-05 14-47-56

There is style issue causing translations to be listed in /post in an a bit weird way (margin looks bad), but I'm not going to fix it right now:
post-padding 2017-10-05 15-03-51

Also, this change generates bunch of useless pages like `/pt/`, `/pt/tags/`, `/pt/post/` and so on. I think that this is not a blocker.

### DIFF
--- a/themes/ooni/layouts/section/post.html
+++ b/themes/ooni/layouts/section/post.html
@@ -24,6 +24,16 @@
               <p class="title">{{ .Title }}</p>
               <p class="byline">{{ .Date.Format "2 Jan 2006" }}</p>
           </a>
+        {{ if .IsTranslated }}
+          {{/* FIXME: looks rather ugly because of bad margins */}}
+          {{ range .Translations }}
+          <ul>
+              {{/* FIXME: convert that to language params, I've not managed to do that quicly */}}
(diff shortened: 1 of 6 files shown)


themes/ooni/layouts/section/post.html
```diff
@@ -24,6 +24,16 @@
               <p class="title">{{ .Title }}</p>
               <p class="byline">{{ .Date.Format "2 Jan 2006" }}</p>
           </a>
+        {{ if .IsTranslated }}
+          {{/* FIXME: looks rather ugly because of bad margins */}}
+          {{ range .Translations }}
+          <ul>
+              {{/* FIXME: convert that to language params, I've not managed to do that quicly */}}
+              <a href="{{ .Permalink }}"><li><p class="title">{{ if eq .Lang "en" }}English{{ else if eq .Lang "pt" }}Portuguese{{ else if eq .Lang "fr" }}French{{ else if eq .Lang "es"  }}Spanish{{ else }}{{ .Lang }}{{ end }}: {{ .Title }}</p></li></a>
```
hellais: I would put this in parens with just the language name after the title

darkk: Maybe. PRs are welcome :)

PR #160, opened 2017-10-05
Add meta-redirect hack to make Catalan translation "first-class citizen"

darkk:
Seems, hugo website generator has no better options.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /public
+/make.conf
 .DS_Store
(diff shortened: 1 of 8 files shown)


hellais:
We should not be automatically redirecting users of our website to some third party site. If we want to include this on our site, we should mirror their page. It's already in markdown: https://github.com/hermes-diactoros/cripto.cat.grav/blob/master/pages/02.blog/censura-internet-referendum/post.md

darkk:
Makes sense.
xref hermes-diactoros/cripto.cat.grav#1

PR #167, opened 2017-10-21
Document website publishing using Travis CI to build it

### LINKED ISSUE #165: Chores: cleanup old OONI webs
* Update the Readme to mention the new deployment procedure
* Remove all references to ooni.github.io
* Remove all the dead code from `Makefile`
* Mention [the diff](https://github.com/TheTorProject/ooni-web/compare/b9d91da4e6658edad838445d13ff6b2c8bb76f17...1c15f7186ca605b3b9716d7fe3ff4b5008955df0) so people can review it
* Mention in the Readme (if there is anything required) what people should do to update their workflow to deploy with the new procedure
* Fix local build

darkk:
That closes #165

Another part of this PR are commits that are already in `master`: https://github.com/TheTorProject/ooni-web/compare/b9d91da4e6658edad838445d13ff6b2c8bb76f17...1c15f7186ca605b3b9716d7fe3ff4b5008955df0


.travis.yml
```diff
@@ -26,6 +26,7 @@ install:
   - git clone --depth 1 --branch master https://github.com/TheTorProject/ooni-probe.git/ contrib/ooni-probe
 
 script:
+  # keep this in-sync with `Makefile`
```
hellais: Why don't you run the makefile from travis?

hellais:
This overall looks good to me. I will look at it more carefully and test it properly, tomorrow.

.travis.yml
darkk: Travis records execution time for each line of the `script`, that's nice to look at these numbers while optimizing build time.

darkk:
@hellais OK. Don't forget to drop https://github.com/OpenObservatory/ooni.github.io repo when this PR is merged.

.travis.yml
```diff
@@ -35,6 +36,8 @@ script:
 
 env:
   # GITHUB_TOKEN=... for @OONIWebPusher see ooni-sysadmin/scripts/mk-travis-vault-env for example
+  # Travis `pages` deploy provides requires "personal access token", that's why it's not a deploy key, see https://docs.travis-ci.com/user/deployment/pages/ for details
+  # Password for the @OONIWebPusher is maintained by leonid@openobservatory.org, password reset email address is admin@openobservatory.org
   - secure: "c9wPMEIZ6qWZi2uQl8RvDq9bVnOhxwDlZPJLyT4m+5WBJau36pBVfZdG8Wxp20dV778jzTrwE1TzFAD80+E95XN0dzQtG2TZASJ/q0IXi7qR4xEnuUi0kDkLAqcvyS1PxQi1W0aUp0t4jYIiqLBDfPol0rJYwTIbsECMv3Rhvyo="
 
 deploy:
```
hellais: > # XXX: wtf does it mean?

It means that it will run the deploy hook when something in merged into the branch named here, in this case `master.

Makefile
```diff
@@ -1,30 +1,25 @@
...
 	touch public/.nojekyll
-	ghp-import public
-	git push -u origin gh-pages:gh-pages --force
-	git push -u ghpage gh-pages:master --force
-	echo "You can now view the website at: http://ooni.github.io/"
-
-# backward compat
-publish-update: update-site
+	cd public && git init && git remote add pages git@github.com:OpenObservatory/openobservatory.github.io.git
+	cd public && git add . && git commit -m 'Manual publishing'
```
hellais: Can we make it so commit history is preserved. I think it's actually a nice feature to have, to be able to see how the website used to be at a given point in time in the past.

It should not be too hard to achieve (basically you just have to clone the repo instead of initing it). I would still do a force push.

I don't think it's important to preserve history before the time in which this is created (though maybe it's not too much effort to just initialise it from the ooni.github.io repo).

hellais:
Another thing to keep in mind is that the design repo is also publishing to openobservatory.github.io/design.

I doesn't seem to create conflicts, but we should either deploy that to it's own domain (design.ooni.io), but in the meantime maybe write a note somewhere about not creating a design folder to avoid conflicts.

I would say a checklist for merging this and considering it done is:

* [ ] Ensure history is preserved and not override the history from scatch
* [ ] Clean up my home directory of staticiforme and ensure I can build the site properly
* [ ] Delete ooni.github.io once history is integrated
* [ ] Ensure that openobservatory.github.io/design works properly

hellais:
Ok I now understand why it was chosen to go for overwriting history. The blocking issues on the travis side seems to be: https://github.com/travis-ci/travis-ci/issues/7562.

I don't actually care enough about preserving history to warrant the effort of adding support for it in a non-native way. I guess we can go ahead with this approach.

hellais:
Ok I reviewed this and checked that everything works. I also move everything inside of /home/ooni to attic except for the `website` dir (the update-website.sh script is unhappy when that dir does not exist).

I am going to merge this.

Makefile
darkk: > Can we make it so commit history is preserved

Nope. I mean, it's possible think to do and nice thing to have, but Travis `pages` provider will drop everything during deploy. It's possible to do deploy manually with another script, but it'll probably increase deployment time a bit. I don't consider that important enough, we can always build older version of website using older hugo binary and older `ooni-web` checkout.

darkk:
> maybe write a note somewhere about not creating a design folder to avoid conflicts

I'd just add something like

```diff
-rm -rf public
+rm -rf public design
```

with proper comment.

PR #168, opened 2017-10-22
Added Amharic translation to Ethiopia report

agrabeli:


darkk:
I can't understand a word, but I'd suggest to give some credits to the translator and/or translating community to make it clear that OONI Team doesn't write amharic (unless anyone does).

hellais:
I wonder if a word better than በኢትዮጵያ could be used, but all in all it looks like a solid translation.

It's good to merge for me.

Tafiti:
@Hellais, what do you mean by ‘better than በኢትዮጵያ could be used’?

Tafiti:
You can credit the translation to ‘Wolete Mariam, Ethiopian translator’.

PR #169, opened 2017-10-23
Make cypherpunks happy again (WRT Google)

darkk:
AFAIK, most of OONI sources use gmail themselves, so moving OONI away from G Suite is not that practical, but [cypherpunks raise good concern](https://lists.torproject.org/pipermail/tor-project/2017-October/001502.html) that there should be some google-proof way to communicate with OONI team for people having Google or US government as an adversary according to their threat model.

IRC is one of possible channels, but IRC has no well-established OTR support. I think, that self-hosted [@torproject.org XMPP server](https://trac.torproject.org/projects/tor/wiki/org/operations/Infrastructure/Jabber) (that explicitly annoys people who don't use OTR) is a good fallback plan for alike sources.

### DIFF
--- a/themes/ooni/layouts/section/about.html
+++ b/themes/ooni/layouts/section/about.html
@@ -42,6 +42,7 @@
     <strong><a href="/about/risks/">documentation about potential risks</a></strong>.
     </p>
 
+    <a name="contact"></a>
     <h2>Contact</h2>
     <p>
     You can contact the OONI team by sending an email to <strong>contact@openobservatory.org</strong>.</p> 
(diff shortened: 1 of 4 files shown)


agrabeli:
Agree 100%.

themes/ooni/layouts/section/about.html
```diff
@@ -54,6 +55,18 @@
...
+    <p>If you're uncomfortable with using email for any of reasons (e.g. because of
+    <a href="https://commons.wikimedia.org/wiki/File:Prism-slide-8.jpg">PRISM</a>
+    watching @openobservatory.org mailbox being hosted by
+    <a href="https://gsuite.google.com" rel="nofollow">G Suite</a>),
+    you can contact team members using XMPP with OTR (when people are online):
+    <ul>
+        <li>Maria Xynou, <a href="xmpp:agrabeli@torproject.org">agrabeli@torproject.org</a>: <tt>97808AA2 2DDBEB34 B661871A 7DC5F86C FAE60873</tt></li>
+        <li>Vasilis Ververis, <a href="xmpp:andz@torproject.org">andz@torproject.org</a>: <tt>23771DC4 585B5C3C 3AC28AA7 D9CD0AE1 60AD8A04</tt></li>
+        <li>Arturo Filastò, <a href="xmpp:art@torproject.org">art@torproject.org</a>: <tt>A806A296 75BE770E 3FF36276 BD5B6CF5 A2BD5DE2</tt></li>
+        <li>Leonid Evdokimov, <a href="xmpp:darkk@torproject.org">darkk@torproject.org</a>: <tt>EC4C71CE 0E2E2221 B746D75B 60533F57 816DD5EB</tt></li>
```
darkk: @hellais @agrabeli @anadahz please verify that fingerprint is correct. I grabbed them from my local OTR fingerprint database, but I've never formally verified them.

agrabeli: I confirm that my fingerprint is correct.

anadahz: My FP is correct.
Thanks @darkk

themes/ooni/layouts/section/about.html
```diff
@@ -54,6 +55,18 @@
 uid       [ultimate] OONI - Open Observatory of Network Interference<br>
 sub   4096R/8EBD2087374399AB 2016-03-23 [expires: 2018-03-26]<br>
     </p>
+
+    <p>If you're uncomfortable with using email for any of reasons (e.g. because of
+    <a href="https://commons.wikimedia.org/wiki/File:Prism-slide-8.jpg">PRISM</a>
+    watching @openobservatory.org mailbox being hosted by
+    <a href="https://gsuite.google.com" rel="nofollow">G Suite</a>),
+    you can contact team members using XMPP with OTR (when people are online):
```
hellais: I would not write it as this. I would simply say:

"Some OONI team can also be reached on jabber at the following addresses"

darkk: _explicit is better than implicit_ :-)

hellais:
LGTM

agrabeli:
I've made some small edits. Merging and publishing.

PR #173, opened 2017-10-30
Fix some missing spaces in the `modern-desktop-app` post

darkk:
:-)


hellais:
Thanks! This was due to a bug in my gdocs to markdown converter: https://github.com/hellais/gdocs-md

PR #174, opened 2017-11-10
FAQ: scrub PII from published data

darkk:
@agrabeli @hellais is it accurate? is it too scary?


agrabeli:
@darkk This is not too scary imo, and I think we should be including this. However, given that this FAQ is now out of date, I am re-writing it. I'll make a pull request with the new FAQ soonish, where I'll also include this point.

darkk:
Okay, I'll just publish it then to keep it in commit history.

PR #176, opened 2017-11-22
Draft of the blog post on UPX & pernambuco.com case

darkk:
I've also cleaned up some tags. That should be, probably, a separate PR, but... let's say I'm just lazy :)

I'd appreciate if someone takes a look at the text.
Feel free to fix typos commiting directly to the branch instead of commenting (or do whatever you consider the right & comfortable thing to do).

PR #177, opened 2017-11-23
Draft edit of the of the blog post on UPX & pernambuco.com case

anadahz:


darkk:
Thanks! I've integrated it to #176.
Yep, #176 has extra irrelevant commit about tags, but having two diverged PRs to discuss same article is probably worse :)

PR #181, opened 2017-12-08
Change URLs to point to the new API

anadahz:


hellais:
Humm, I am not sure we should be changing the links inside of the old blog posts. Also the change should not be just a search and replace of the string, it should also change the text to call it the OONI API (instead of Measurements interface).

Moreover on the data page we should update the logo too, so that it shows the OONI API logo.

anadahz:
I think it's good to update the old blog posts URLs as visitors and crawlers will follow and index links accordingly.

hellais:
> I think it's good to update the old blog posts URLs as visitors and crawlers will follow and index links accordingly.

We should actually do that by setting up a 301 redirect from measurements.ooni.io -> api.ooni.io

anadahz:
Reverted the URL changes to blog posts.
We can add the OONI API image when @elioqoshi approves the/a logo.

hellais:
> We can add the OONI API image when @elioqoshi approves the/a logo.

I already added the logos and made some changes to the markup in: https://github.com/TheTorProject/ooni-web/pull/182.

@anadahz Can you close this PR to avoid confusion?

anadahz:
> I already added the logos and made some changes to the markup in: #182.
@anadahz Can you close this PR to avoid confusion?

Sure, it seems that you forgot to change some text and URL.
Compare this PR with #182 and when it's OK close this PR.

hellais:
> Sure, it seems that you forgot to change some text and URL.

If you mean in the faq, that is anyways being re-written and is not linked from the homepage. So I am not sure what you mean.

anadahz:
>If you mean in the faq, that is anyways being re-written and is not linked from the homepage.

Indeed I was talking about the FAQ: https://ooni.torproject.org/faq/faq/

PR #182, opened 2017-12-08
Replace links with OONI API and make minor copy edits

hellais:
@elioqoshi do you think we should be using the monochrome or the color version of the logo?

For the moment I committed both, but let me know what you think is best.

screen shot 2017-12-08 at 20 28 45
screen shot 2017-12-08 at 20 40 04

I feel like the colored one is a bit to "in your face" and it's probably best to use the monochrome version.


anadahz:
@hellais it seems that you forgot to rebuild the OONI website for the changes to take effect.

Please do build the website or throw a message/ping someone so that the website will always be in sync with the master branch.

hellais:
I did actually ping people in slack about it. I will rebuild it tomorrow.

--  
Arturo Filastò

hellais:
Published.

anadahz:
Great thanks!

PR #189, opened 2018-01-05
Iran Protests: OONI data confirms censorship events

darkk:
@agrabeli here is markdown version of the document you sent me, but it misses the image on WhatsApp.

I hope, you have a source of that image as I have not managed to get reasonable (non-blurred) version of the image out of our wonderful proprietary collaboration platform :-)


agrabeli:
I think it's ready to merge and publish.

PR #190, opened 2018-01-12
Add subscribe for the net neutrality in the US campaign

hellais:
This is related to the FFTF campaign.

### DIFF
--- a/content/get-involved/net-neutrality-us.md
+++ b/content/get-involved/net-neutrality-us.md
@@ -1,5 +1,5 @@
 Add your email to receive updates about Net Neutrality in the US.
 
-{{<mailchimp-form nonce="b_7ef399cbb419111525ff38fda_c25f83d1ab" action="https://ooni.us17.list-manage.com/subscribe/post?u=7ef399cbb419111525ff38fda&amp;id=c25f83d1ab" >}}
+{{<mailchimp-form spamid="b_7ef399cbb419111525ff38fda_c25f83d1ab" action="https://ooni.us17.list-manage.com/subscribe/post?u=7ef399cbb419111525ff38fda&amp;id=c25f83d1ab" >}}
 
 We promise we will not spam you.
--- a/content/get-involved/net-neutrality-us.md
+++ b/content/get-involved/net-neutrality-us.md
@@ -1,0 +1,5 @@
+Add your email to receive updates about Net Neutrality in the US.
+
+{{<mailchimp-form nonce="b_7ef399cbb419111525ff38fda_c25f83d1ab" action="https://ooni.us17.list-manage.com/subscribe/post?u=7ef399cbb419111525ff38fda&amp;id=c25f83d1ab" >}}
+
+We promise we will not spam you.
--- a/layouts/shortcodes/mailchimp-form.html
+++ b/layouts/shortcodes/mailchimp-form.html
@@ -10,7 +10,7 @@
 	
 	<input type="email" value="" name="EMAIL" class="email" id="mce-EMAIL" placeholder="email address" required>
     <!-- real people should not fill this in and expect good things - do not remove this or risk form bot signups-->
-    <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="{{ .Get "nonce" }}" tabindex="-1" value=""></div>
+    <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="{{ .Get "spamid" }}" tabindex="-1" value=""></div>
     <div class="clear"><input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="button"></div>
     </div>
 </form>
--- a/layouts/shortcodes/mailchimp-form.html
+++ b/layouts/shortcodes/mailchimp-form.html
@@ -1,0 +1,17 @@
+<link href="//cdn-images.mailchimp.com/embedcode/slim-10_7.css" rel="stylesheet" type="text/css">
+<style type="text/css">
+	#mc_embed_signup{background:#fff; clear:left; font:14px Helvetica,Arial,sans-serif; }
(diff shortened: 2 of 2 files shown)


content/get-involved/net-neutrality-us.md
```diff
@@ -0,0 +1,5 @@
+Add your email to receive updates about Net Neutrality in the US.
+
+{{<mailchimp-form nonce="b_7ef399cbb419111525ff38fda_c25f83d1ab" action="https://ooni.us17.list-manage.com/subscribe/post?u=7ef399cbb419111525ff38fda&amp;id=c25f83d1ab" >}}
```
darkk: Does it really work that way?
Hardcoded value that is called `nonce` and is used over and over again looks really strange to me...

hellais: It's not really a nonce, in the sense that is actually used multiple times, though it's unique for every "list" you create. It does work as I have tested that it will allow me to subscribe to the list.

hellais: Maybe `spam_prevent_id` is a better value for it. The purpose of it is if automatic bots submit the form with that value then mailchimp will not consider the subscription.

bassosimone: How does this work exactly?

hellais: @bassosimone https://gohugo.io/content-management/shortcodes/

hellais:
> How much did you customize the code compared to the code generated by the mailchimp form builder?

I did no customisation whatsoever. I just factored the variable stuff into a hugo shortcode.

content/get-involved/net-neutrality-us.md
```diff
@@ -0,0 +1,5 @@
+Add your email to receive updates about Net Neutrality in the US.
+
+{{<mailchimp-form spamid="b_7ef399cbb419111525ff38fda_c25f83d1ab" action="https://ooni.us17.list-manage.com/subscribe/post?u=7ef399cbb419111525ff38fda&amp;id=c25f83d1ab" >}}
```
bassosimone: @hellais so, this is the shortcode. It is somehow similar to a React "class", IIUC.

bassosimone: Like, it has parameters that you can set, etc.

```JavaScript
<MailchimpForm
     spamid="....."
     action="..."
/>
```

This above would be the equivalent in React.

Right?

layouts/shortcodes/mailchimp-form.html
```diff
@@ -0,0 +1,17 @@
+<link href="//cdn-images.mailchimp.com/embedcode/slim-10_7.css" rel="stylesheet" type="text/css">
+<style type="text/css">
+	#mc_embed_signup{background:#fff; clear:left; font:14px Helvetica,Arial,sans-serif; }
+	/* Add your own MailChimp form style overrides in your site stylesheet or in this style block.
+	   We recommend moving this block and the preceding CSS link to the HEAD of your HTML file. */
+</style>
+<div id="mc_embed_signup">
+  <form action="{{ .Get "action" }}" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate" target="_blank" novalidate>
```
bassosimone: Is it correct to state that `"{{ .Get "action" }}" will be replaced with the value of the `action` parameter passed to the shortcode? @hellais

bassosimone: Also, can you wrap this line?

bassosimone: Also, I am confused by `class=validate` and then `novalidate`. I assume this is the mailchimp template but I do not fully understand the implication. Does it mean that validation is only done in JavaScript?

layouts/shortcodes/mailchimp-form.html
```diff
@@ -0,0 +1,17 @@
...
+    <div id="mc_embed_signup_scroll">
+	
+	<input type="email" value="" name="EMAIL" class="email" id="mce-EMAIL" placeholder="email address" required>
+    <!-- real people should not fill this in and expect good things - do not remove this or risk form bot signups-->
+    <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="{{ .Get "spamid" }}" tabindex="-1" value=""></div>
+    <div class="clear"><input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="button"></div>
```
bassosimone: Can you also wrap the chunk above?

content/get-involved/net-neutrality-us.md
hellais: I guess you can say that.

layouts/shortcodes/mailchimp-form.html
hellais: I have no idea how this works. As I said above, I just copy pasted this verbatim from the mailchimp web interface and replaced some strings with variables.

layouts/shortcodes/mailchimp-form.html
hellais: I would prefer not to wrap it as this is how it's presented in the mailchimp web interface and it makes it easier to ensure it matches 100% what is written in there and/or replace it with a new version.

bassosimone: Good point!

hellais:
> I still do not get how embedding into the page a specific code helps to prevent spam.
> Is it because the bot must be slightly smarter and must also make sure that the hidden input is submitted?

I think that is the case. I don't know exactly how this works, I just copy pasted the snippet from the mailchimp web interface and verified it works.

hellais:
I am going to merge and deploy this.

layouts/shortcodes/mailchimp-form.html
bassosimone: Ok

PR #194, opened 2018-02-03
Add post on OONI data availability

darkk:
@agrabeli I don't expect you to be able to review the post given weekend, ongoing vacation and other sad circumstances, but here is it if you want to spend some time reading it before I hit `publish` button this evening.


darkk:
There is no more urgency in publishing that as the person I consider to be the main recipient of the message has already seen it according to the @hellais :)

agrabeli:
@darkk Thanks for writing this post - I think many researchers can benefit from this. :)

I have provided some suggested edits in 2 commits. Otherwise, I think it's ready for publication.

PR #216, opened 2018-07-02
replaced Egypt Report Blog image with updated ones

elioqoshi:
these images have a bigger pattern and adjusted illustration


hellais:
Based against the wrong branch. Closing as it's already integrated into the correct branch.

PR #232, opened 2018-09-27
Fix ooni-%20probe link & make ooni.io links relative

darkk:


.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
   - tar -xf /tmp/hugo_0.29_Linux-64bit.tar.gz -C bin/
 
   - pip install sphinx sphinx_rtd_theme # Should I specify sphinx versions?
-  - git clone --depth 1 --branch master https://github.com/TheTorProject/ooni-probe.git/ contrib/ooni-probe
```
hellais: Does this really not work?

Because I had made it so the github redirects should make it possible to keep cloning probe-legacy from thetorproject/ooni-probe path.

darkk: It should and, seems, it does.
I was trying to understand the reason for Travis build to fail and just updated the URL during that process :)

darkk:
Fixed build in https://github.com/ooni/probe-legacy/pull/42

PR #233, opened 2018-10-02
VE: add section on Tor unblocking

darkk:
@andresAzp, here is a short note summarizing what I know regarding Tor unblocking in VE.

Feel free to add details if you have time and will to collect them.

Unfortunately, most of `vanilla_tor` measurements coming from OONI Probe daily are corrupted, so I failed to say anything on top of that missing data :-(

**TODO**: add following comment to https://trac.torproject.org/projects/tor/ticket/26807 when that's merged.
> Here are some parts of the story, captured by OONI:
> https://ooni.torproject.org/post/venezuela-internet-censorship/#blocking-of-tor
> Seems, Tor was unblocked in Venezuela a month ago. Is it relatively close to the truth?


darkk:
I suggest to update "update date" date and merge it. If additional information will be found then we'll publish another update.

PR #234, opened 2018-10-03
Add BTC account for donations

darkk:
Serving it from imgur is "insecure" and also triggers mixed content warning from netlify.


bassosimone:
👍

bassosimone:
Oh, wait, maybe I'm misled. Shouldn't you also change some URL somewhere?

PR #245, opened 2019-01-09
Fix (false) mixed content warnings from Netlify

darkk:
Hopefully, it fixes the _"Alert: Insecure mixed content detected in ooni"_.


darkk:
Seems, it does.

Issue #258, opened 2019-03-29
GPG key error


xhdix:
To install OONI in Debian, there is a problem getting GPG key.
https://github.com/TheTorProject/ooni-web/blob/master/content/install/ooniprobe.md

```
root@dbn:/home/user# gpg --keyserver keys.gnupg.net --recv A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
gpg: keyserver receive failed: No keyserver available
root@dbn:/home/user# gpg --keyserver keys.gnupg.net --recv A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
gpg: keyserver receive failed: No keyserver available
root@dbn:/home/user# apt-key adv --keyserver keys.gnupg.net --recv-key A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89     
Executing: /tmp/apt-key-gpghome.6FwcbNgQd8/gpg.1.sh --keyserver keys.gnupg.net --recv-key A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
gpg: keyserver receive failed: No keyserver available
root@dbn:/home/user# apt-key adv --keyserver hkp://keys.gnupg.net --recv-key A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
Executing: /tmp/apt-key-gpghome.KTXyWIAuiM/gpg.1.sh --keyserver hkp://keys.gnupg.net --recv-key A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
gpg: keyserver receive failed: No keyserver available
```

Previously, the solution to this problem was to use the following commands:

```
curl https://deb.torproject.org/torproject.org/A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89.asc | gpg --import
​
gpg --export A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89 | apt-key add - 
```

Result:

```
root@dbn:/home/user# curl https://deb.torproject.org/torproject.org/A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89.asc | gpg --import
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 19665  100 19665    0     0   4702      0  0:00:04  0:00:04 --:--:--  5011
gpg: key EE8CBC9E886DDD89: 36 signatures not checked due to missing keys
[...]

PR #259, opened 2019-04-01
Fix PGP/GPG Error

xhdix:
according to https://github.com/TheTorProject/ooni-web/issues/258


hellais:
Thanks for looking at this. Merging!

PR #262, opened 2019-04-12
Minor edit to VE report

agrabeli:
I have removed Alp Toker from the authors of this report because he did not contribute to it.


hellais:
lgtm.

PR #265, opened 2019-05-04
Uganda: add traceroute to 25/SMTP TCP port snippet

darkk:
I've stumbled upon an interesting datapoint from Uganda while cleaning up various notes.

The datapoint shows that tested Uganda network had two distinct filters at least:
- blocking tcp/80 and tcp/443 towards IP addresses of twitter.com
- blocking tcp/25 to prevent SPAM

Of course it could also be the same filter with different settings regarding IP.TTL field processing, it's something that this technique is unable to distinguish.

I want to add that snippet to refer to it from ooni/spec as an another kinda interesting example for traceroute technique.


hellais:
lgtm

PR #267, opened 2019-05-05
 Add translation of post on wikipedia blocking 

onoketa:


hellais:

PR #278, opened 2019-08-12
Edited Web Connectivity test description

agrabeli:


agrabeli:
@bassosimone wow, thanks for the fast review! :) Merging and pushing an update to the site.

PR #298, opened 2019-10-26
OONI Glossary

agrabeli:
Special thanks to Anatol for contributing to this glossary! :)

### DIFF
--- a/content/about/glossary.md
+++ b/content/about/glossary.md
@@ -48,7 +48,7 @@ glossary.
 
 * [DPI](#dpi)
 
-* [False positive](#false positive)
+* [False positive](#false-positive)
 
 * [HTTP](#http)
 
@@ -282,7 +282,7 @@ DNS stands for "Domain Name System" and it maps domain names to IP addresses.
 
 A domain is a name that we attribute to websites (when we create them), so that
 we can more easily remember and access them. For example, twitter.com is the
-domain of the Twitter website (https://twitter.com/).
+domain of the Twitter website (`https://twitter.com/`).
 
 However, computers can't connect to internet services through domain names, but
 based on IP addresses: the digital address of each service on the internet.
@@ -312,8 +312,8 @@ deliberately been blocked.
 exist or has been misspelled.
 
 When Internet Service Providers (ISPs) receive government orders to block
-specific websites, they sometimes adopt this technique. In these cases, **it may
-be possible to circumvent the censorship by merely changing your DNS resolver**.
+specific websites, they sometimes adopt this technique. In these cases, it may
+be possible to circumvent the censorship by merely changing your DNS resolver.
 
 ### DNS lookup
 
@@ -396,8 +396,8 @@ The [OONI Probe apps](https://ooni.org/install/) measure the [DNS tampering of w
 A domain is a name that we attribute to websites (when we create them), so that
 we can more easily remember and access them.
 
-For example, twitter.com is the domain of the Twitter website
-(https://twitter.com/).
+For example, **twitter.com** is the domain of the Twitter website
+(`https://twitter.com/`).
 
 ### DPI
 
@@ -461,7 +461,7 @@ The HTTP protocol allows communication between a client and a server. It does so
 by handling a client’s request to connect to a server, and a server’s response
 to a client’s request.
 
-All websites include an HTTP (or HTTPS) prefix (such as http://example.com/) so
+All websites include an HTTP (or HTTPS) prefix (such as `http://example.com/`) so
 that your computer (the client) can request and receive the content of a website
 (hosted on a server).
 
@@ -473,7 +473,7 @@ Hypertext Transfer Protocol Secure (HTTPS) -- also known as HTTP over TLS, or
 HTTP over SSL --  is the **encrypted version of the HTTP protocol**.
 
 Over the last years, many websites on the internet started supporting HTTPS
-(such as https://www.facebook.com/) so that the transmission of data (such as
+(such as `https://www.facebook.com/`) so that the transmission of data (such as
 passwords to login to websites) over the HTTP protocol is encrypted.
 
 ### HTTP blocking
@@ -563,20 +563,20 @@ HTTP response.
 
 Common HTTP status codes include:
 
-* HTTP Error **500*(Internal Server Error): General-purpose error message when a
+* HTTP Error *500*(Internal Server Error): General-purpose error message when a
 web server encounters some form of internal error.
 
-* HTTP Error **403*(Forbidden): When you try to access a forbidden directory on
+* HTTP Error *403*(Forbidden): When you try to access a forbidden directory on
 a website.
 
-* HTTP Error **404*(Not Found): When you try to access a resource on a web
+* HTTP Error *404*(Not Found): When you try to access a resource on a web
 server (usually a web page) that doesn’t exist (because, for example, you have
 mistyped a URL).
 
-* HTTP Error **400*(Bad Request): When the application you are using (such as
+* HTTP Error *400*(Bad Request): When the application you are using (such as
 your web browser) accesses a server incorrectly or the request was corrupted.
 
-* HTTP Error **401*(Unauthorized): When you try to access a restricted web page
+* HTTP Error *401*(Unauthorized): When you try to access a restricted web page
 but aren't authorized to do so (usually because of a failed login attempt).
 
 When measuring the blocking of websites, the OONI Probe [Web Connectivity test](https://ooni.org/nettest/web-connectivity/) compares the HTTP status codes
@@ -634,8 +634,7 @@ case, it constitutes a form of internet censorship -- or it may have been caused
 Since [OONI Probe](https://ooni.org/install/) requires internet connectivity in order to perform tests,
 measuring internet blackouts is currently out of scope.
 
-Several [public data sources](https://ooni.org/post/examining-internet-
-blackouts/) are available for monitoring internet blackouts, such as
+Several [public data sources](https://ooni.org/post/examining-internet-blackouts/) are available for monitoring internet blackouts, such as
 [IODA](https://ioda.caida.org/).
 
 ### Internet censorship
@@ -687,8 +686,7 @@ Each [OONI Probe test](https://ooni.org/nettest/) is designed to measure
 different forms of network interference. Depending on the test, each OONI
 measurement contains a different result.
 
-Here is an example of an OONI measurement: https://explorer.ooni.org/measurement/20191024T090549Z_AS42668_D8RVKYKWbqzaTopEcH3K6qy8yzOHVe1QgNeyzn686G1CbEveKU?inp
-ut=http://kavkaznews.com/
+Here is an example of an OONI measurement: https://explorer.ooni.org/measurement/20191024T090549Z_AS42668_D8RVKYKWbqzaTopEcH3K6qy8yzOHVe1QgNeyzn686G1CbEveKU?input=http://kavkaznews.com/
 
 ### Metadata
 
@@ -772,8 +770,8 @@ increased privacy and security benefits.
 Websites, for example, can be placed in an anonymous network location through
 the use of onion services, which allow them to hide their IP address and provide
 more privacy to their visitors. You can distinguish such websites through their
-.onion addresses, such as https://facebookcorewwwi.onion/ or
-http://www.nytimes3xbfgragh.onion/.
+.onion addresses, such as `https://facebookcorewwwi.onion/` or
+`http://www.nytimes3xbfgragh.onion/`.
 
 Over the last years, the [OONI Probe version for Linux and macOS](https://ooni.org/install/ooniprobe/) used onion services to upload
 measurements to OONI servers.
@@ -938,7 +936,7 @@ cryptographic protocol that allows you to maintain a secure, encrypted
 connection between your computer and an internet service.
 
 When you connect to a website though TLS, the address of the website will begin
-with HTTPS (such as https://www.facebook.com/), instead of HTTP.
+with HTTPS (such as `https://www.facebook.com/`), instead of HTTP.
 
 ### Tor
 
@@ -961,7 +959,7 @@ manipulate internet traffic.
 
 A URL is the address of a World Wide Web page.
 
-For example, https://twitter.com/ is a URL, while twitter.com is a domain.
+For example, `https://twitter.com/` is a URL, while twitter.com is a domain.
 
 ### VPN
 
--- a/content/about/glossary.md
+++ b/content/about/glossary.md
@@ -3,10 +3,7 @@
 This glossary contains *brief explanations* for terms used in OONI apps,
 methodologies, and research reports.
 
-If you would like to see additional terms in this glossary, please [open a pull request](https://github.com/TheTorProject/ooni-web) or [send us an email](https://ooni.org/about/#contact).
-
-Special thanks to Anatol (OONI community member) for contributing to this
-glossary.
+If you would like to see additional terms in this glossary, please [open a pull request](https://github.com/ooni/ooni.org) or [send us an email](https://ooni.org/about/#contact).
 
 * [API](#api)
 
@@ -26,8 +23,6 @@ glossary.
 
 * [Client](#client)
 
-* [Cloud-fronting](#cloud-fronting)
-
 * [Data processing pipeline](#data-processing-pipeline)
 
 * [DNS](#dns)
@@ -44,7 +39,9 @@ glossary.
 
 * [DNS tampering](#dns-tampering)
 
-* [Domain](#domain)
+* [Domain fronting](#domain-fronting)
+
+* [Domain name](#domain-name)
 
 * [DPI](#dpi)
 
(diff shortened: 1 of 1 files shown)


content/about/glossary.md
```diff
@@ -0,0 +1,973 @@
...
+* [VPN](#vpn)
+
+### API
+
+An Application Programming Interface (API) is an interface that enables
+programmers to write integration code between different services.
+
+The [OONI API](https://api.ooni.io/) provides an interface that enables users to
+request specific types of OONI data and to download such data in a specific
+format (JSON).
```
hellais: We should mention in here that the OONI API is not suitable for batch usage and point people instead to the MetaDB.

content/about/glossary.md
```diff
@@ -0,0 +1,973 @@
...
+Each AS is responsible for delivering IP packets to a set of IP addresses that
+it manages.
+
+### ASN
+
+An Autonomous System Number (ASN) is a unique identifier of an autonomous system
+(AS). This number allows its autonomous system to exchange routing information
+with other systems.
+
+Internet Service Providers (ISPs) must have an officially registered ASN.
```
hellais: I am not sure this is actually a legal requirement. It's probably desirable for them to have one, but AFAIK you can be a ISP without having an ASN. It's also crucial I think to say that ISPs can have also more than one ASN.

content/about/glossary.md
```diff
@@ -0,0 +1,973 @@
+# OONI Glossary
+
+This glossary contains *brief explanations* for terms used in OONI apps,
+methodologies, and research reports.
+
+If you would like to see additional terms in this glossary, please [open a pull request](https://github.com/TheTorProject/ooni-web) or [send us an email](https://ooni.org/about/#contact).
```
hellais: The PR link needs updating the repo is not `ooni/ooni.org`.

content/about/glossary.md
```diff
@@ -0,0 +1,973 @@
...
+
+A simple Google search of an ASN number will show you which ISP it corresponds
+to. For example, if you search "AS30722", you will see that it belongs to
+"Vodafone Italia".
+
+### BGP
+
+The Border Gateway Protocol (BGP) is a standardized exterior gateway protocol
+designed to exchange routing and reachability information among autonomous
+systems (AS) on the internet.
```
hellais: I think if we are going for a simple definition it's not really useful to drop things like "standardized exterior gateway protocol
designed to exchange routing and reachability information among autonomous
systems (AS) on the internet".

I would say something along the lines of:

The Border Gateway Protocol (BGP) is a protocol used by Autonomous Systems (AS) on the internet to exchange routing information. With this information they are able to discover which AS is responsible for which set of IP address and what is the path IP packets need to take in order to reach that particular AS.

content/about/glossary.md
bassosimone: My understanding is that you need an ASN to use BGP. It may be that you can be an ISP without BGP, but I think it seldom happens in practice, so I guess we should say that an ISP has one or more ASN.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+
+### Client
+
+In the world of computers, a client is a piece of software or hardware that
+accesses a service made available by a server.
+
+The [OONI Probe app](https://ooni.org/install/), for example, is a client and it
+communicates with OONI servers in order to submit testing results.
+
+### Cloud-fronting
```
hellais: I suggest renaming this to Domain Fronting. Cloud fronting is derived from the AWS product called "AWS Cloud Front", while Domain Fronting is the non vendor specific term.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+
+Cloud-fronting (commonly referred to as "domain-fronting") is a technique that
+circumvents internet censorship by appearing to connect to the domain of a cloud
+provider  -- such as google.com or awsstatic.com (Amazon cloud infrastructure)
+-- while actually connecting to a blocked website.
+
+This technique, therefore, not only provides users access to blocked internet
+services, but it also hides the fact that they are accessing those services at
+all.
```
hellais: I would refrase this section as follows:

Domain fronting (also sometimes known as "cloud fronting") is a censorship circumvention technique which relies on using a popular domain name hosted on a big cloud service provider to "front" traffic towards a circumvention service hosted on the same cloud service.

For example I may setup on the Google cloud a service under the domain `circumvention-service.google-cloud.com`, but front it via `google.com`. This means that from the perspective of somebody observing network traffic, all that they see is an encrypted TLS connection towards `google.com`, yet inside of this encrypted connection the client then says "actually I want to access `circumvention-service.google-cloud.com`", circumventing any potential block.

This means that the censor would either have to block all of the google cloud (and therefore disrupt many more service too as a result) or they will have a hard time distinguishing requests towards `circumvention-service.google-cloud.com` from requests to `google.com`. This concept take the name of "collateral freedom".

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+responsible for aggregating network measurement data from OONI Probe users and
+analyzing the data to identify network anomalies around the world.
+
+### DNS
+
+DNS stands for "Domain Name System" and it maps domain names to IP addresses.
+
+A domain is a name that we attribute to websites (when we create them), so that
+we can more easily remember and access them. For example, twitter.com is the
+domain of the Twitter website (`https://twitter.com/`).
```
hellais: I would drop the `https://twitter.com` from the parens as that is not the website, but rather the URL.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+address of a particular domain name and your DNS resolver (the server managing
+your DNS requests) is dishonest, intentionally providing you an incorrect
+answer.
+
+As a result, you receive a forged response, such as the following:
+
+* **NXDOMAIN:** A response indicating that the requested domain is either not
+registered or invalid.
+
+* **Block page:** A web page that informs you that the requested domain has
```
hellais: This should say "The IP Address of a blockpage"

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+deliberately been blocked.
+
+* **Error page:** A web page that informs you that the requested domain does not
```
hellais: This is the same as the two mentioned above from the perspective of DNS.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+exist or has been misspelled.
+
+When Internet Service Providers (ISPs) receive government orders to block
+specific websites, they sometimes adopt this technique. In these cases, it may
+be possible to circumvent the censorship by merely changing your DNS resolver.
```
hellais: add "or using encrypted DNS"

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+
+When you perform a DNS lookup, you ask your DNS resolver (the server managing
+your DNS requests) to disclose which IP addresses correspond to specific domain
+names.
+
+To check whether websites are blocked by means of DNS, [OONI Probe](https://ooni.org/install/) starts off by performing a DNS lookup to see
+which IP addresses are mapped to tested domains by the user's DNS resolver. If
+the IP addresses do *not* match those allocated by another DNS resolver (known
+to not implement censorship), it's possible that censorship is implemented at
+the DNS level.
```
hellais: Our test is a bit more complex than this, maybe it would be useful to mention that this is a simplified version of how the web_connectivity test works and link to the test specification for more details?

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+A DNS query (otherwise known as a "DNS request") is a request for information
+sent from a user's computer to a DNS server.
+
+In most cases, a DNS request is sent to ask for the IP address associated with a
+domain name (such as twitter.com).
+
+### DNS resolver
+
+A DNS resolver is a server that stores a central database of DNS nameservers
+(which define the DNS providers of domains), operating like an "address book".
```
hellais: I would rephrase this definition as:

A DNS resolver is a server which maps domain names to IP address, operating like an "address book".

Saying that it's a central database is technically wrong.

Also drop database from below.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+* **DNS spoofing:** Where your DNS request is intercepted and you receive the
+wrong IP address.
+
+DNS tampering is a common censorship technique adopted by Internet Service
+Providers (ISPs) around the world.
+
+The [OONI Probe apps](https://ooni.org/install/) measure the [DNS tampering of websites and apps](https://ooni.org/nettest/).
+
+### Domain
+
```
hellais: This should be "Domain name"

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+A domain is a name that we attribute to websites (when we create them), so that
+we can more easily remember and access them.
+
+For example, **twitter.com** is the domain of the Twitter website
+(`https://twitter.com/`).
+
+### DPI
+
+Deep packet inspection (DPI) is an advanced method of examining and managing
```
hellais: I would drop "advanced method", depending on the type of DPI it canbe advanced or not.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+All websites include an HTTP (or HTTPS) prefix (such as `http://example.com/`) so
+that your computer (the client) can request and receive the content of a website
+(hosted on a server).
+
+The transmission of data over the HTTP protocol is **unencrypted**.
+
+### HTTPS
+
+Hypertext Transfer Protocol Secure (HTTPS) -- also known as HTTP over TLS, or
+HTTP over SSL --  is the **encrypted version of the HTTP protocol**.
```
hellais: I would say rather than:

> encrypted version of the HTTP protocol

I would go for

> the HTTP protocol over an encrypted channel

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+An HTTPS collector is a server that collects data over the encrypted HTTPS
+protocol.
+
+OONI utilizes HTTPS collectors in order to collect network measurements from
+OONI Probe users around the world.
+
+### HTTP header
+
+An HTTP header is the section of the request and response messages that defines
+the operating parameters of an HTTP transaction between a client and a server.
```
hellais: I would rephrase this as:

HTTP headers are used to transmit to the server or the client metadata about the request or response. Certain HTTP headers are standardised for transmitting common bits of information, such as the `User-Agent` header field, which is used to tell the server which browser is performing the request.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+
+When measuring the blocking of websites, the OONI Probe [Web Connectivity test](https://ooni.org/nettest/web-connectivity/) compares the HTTP headers of
+the tested website from the network of the user with a non-censored network (as
+part of a larger testing methodology).
+
+### HTTP request
+
+In order to connect to a server hosting an internet service (such as a website
+or application), you need to request access to it through the HTTP protocol.
+This is what we call an HTTP request.
```
hellais: I would drop this whole section because it's inaccurate.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+website it is hosting.
+
+In some cases though, Internet Service Providers (ISP) prevent users from
+accessing certain websites by blocking or interfering with the connection
+between them and the server. This is a form of HTTP blocking,
+[measured](https://ooni.org/nettest/web-connectivity/) by the [OONI Probe app](https://ooni.org/install/).
+
+### HTTP response
+
+Every time you connect to a website, your browser (the client) sends a request
```
hellais: I would replace "connect" with "visit", because connect has a specific meaning in the context of internet protocols.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+("HTTP request") through the HTTP protocol to the server which is hosting that
+website.
+
+A server normally responds with the content of the website it is hosting. This
```
hellais: I would rephrase this as:

A server will send in response to a HTTP requests a HTTP response, which includes HTTP response headers and optionally a response body, which in the case of a website will be the content of the page.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+
+* HTTP Error *404*(Not Found): When you try to access a resource on a web
+server (usually a web page) that doesn’t exist (because, for example, you have
+mistyped a URL).
+
+* HTTP Error *400*(Bad Request): When the application you are using (such as
+your web browser) accesses a server incorrectly or the request was corrupted.
+
+* HTTP Error *401*(Unauthorized): When you try to access a restricted web page
+but aren't authorized to do so (usually because of a failed login attempt).
```
hellais: You should replace the above with: HTTP status xxx

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+A server normally responds with the content of the website it is hosting. This
+response is an HTTP response.
+
+### HTTP status codes
+
+HTTP status codes are standardized codes (maintained by the Internet Assigned
+Numbers Authority) that are issued by a server in response to a client's
+request. As suggested by their name, these codes communicate the status of an
+HTTP response.
+
```
hellais: I would add something saying:

HTTP status codes are divided into specific ranges, depending on the type of status they are communicating.

HTTP status codes that go from 200-299 are used to indicate a successful request
HTTP status codes that go from 300-399 are used to indicate that there are multiple choices, for example to redirect the request to another site
HTTP status codes that go from 400-499 are used to indicate errors in the request by the client
HTTP status code that go from 500-599 are used to indicate server side errors

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+Common HTTP status codes include:
+
```
hellais: We should also add:

* HTTP Status *200*(OK): Used to indicate that the request was successful. This is generally the status code used most commonly for successful page loads.

* HTTP Status *301*(Moved Permanently): When the resource requested by the client is now available at a different location.

* HTTP Status *302*(Found): When the resource requested by the client can be **temporarily accessed** at a different location.

hellais: Also maybe just for fun, we should add:
* HTTP Status 451 (Unavailable For Legal Reasons): When the resource that is being tried to access is blocked by the server due to some legal restrictions. This is basically the status code to be used for implementing internet censorship.

content/about/glossary.md
agrabeli: Thanks, I've added a sentence to mention this.

content/about/glossary.md
agrabeli: Thanks for the feedback guys. I've edited accordingly.

content/about/glossary.md
agrabeli: Good catch! Edited.

content/about/glossary.md
agrabeli: Thanks! This is very helpful.

content/about/glossary.md
agrabeli: I think we referred to it as "cloud fronting" in our documentation, but I agree that it's probably best to include "domain fronting" in the glossary.

content/about/glossary.md
agrabeli: +1

content/about/glossary.md
agrabeli: This is great, thank you!

content/about/glossary.md
agrabeli: OK

content/about/glossary.md
agrabeli: I've edited this section per your suggestions.

content/about/glossary.md
agrabeli: Thanks! Added.

content/about/glossary.md
agrabeli: OK, I've added a sentence to link to the Web Connectivity test description.

content/about/glossary.md
agrabeli: Thanks, edited.

content/about/glossary.md
agrabeli: OK

content/about/glossary.md
agrabeli: OK

content/about/glossary.md
agrabeli: OK

content/about/glossary.md
agrabeli: Thanks!

content/about/glossary.md
agrabeli: OK. Is the remaining description of "HTTP request" accurate?

content/about/glossary.md
agrabeli: OK

content/about/glossary.md
agrabeli: Thanks!

content/about/glossary.md
agrabeli: Sure

content/about/glossary.md
agrabeli: Thanks, added!

content/about/glossary.md
agrabeli: Thanks, I've added them all.

content/about/glossary.md
```diff
@@ -0,0 +1,975 @@
...
+### HTTP request
+
+In order to connect to a server hosting an internet service (such as a website
+or application), you need to request access to it through the HTTP protocol.
+This is what we call an HTTP request.
+
+Every time you connect to a website, your browser (the client) sends a request
+("HTTP request") through the HTTP protocol to the server which is hosting that
+website. A server normally responds ("HTTP response") with the content of the
+website it is hosting.
```
hellais: I suggest replacing:

> A server normally responds ("HTTP response") with the content of the website it is hosting.

with

> A server normally replies with a "HTTP response" which includes the content of the website it is hosting.

content/about/glossary.md
hellais: Added a small edit below.

agrabeli:
In my latest commit I edit the glossary based on feedback shared by @hellais & @p1g30n. I would appreciate further review to prepare the glossary for publication.

agrabeli:
@hellais My latest commit includes edits based on the last rounds of feedback. If the glossary looks good to you, please merge it.

PR #311, opened 2019-11-29
Add collected data, fix links, mv ooniprobe

anadahz:
* Enumerate default data published in OONI reports
* Fix source code URLs to point to the correct location
* Rename ooniprobe to OONI Probe


hellais:
The data policy has received several updates in the meantime. I am closing this PR.

Issue #314, opened 2019-12-16
Improve SEO of ooni.org


xhdix:
Due to the HUGO limitation I don't know how much SEO can be done.
(It is said that search engine bots are blind. They follow braille(HTML) lines to increase their understanding of each page.)
Here are some things to get started:

**High priority:**

- [x] Change title to: `<title> {{.Title}} - OONI </title>`  (or: `{{if .IsHome}}{{ $.Site.Title }}{{else}}{{.Title}}{{end}}` )
- [x] Each page should have a unique title and description. (Especially in the meta tag -- OpenGraph and Twitter) . First we need to add the `description` value in all the posts beside the `title`. Next, change the `head.html` file as the `title` :point_up_2: .

~Images must have a relevant name. (Not just the name in the image. Google can read the texts in the images itself). Something like `Image-title-Post-title-Time.jpg` or more. (not jpeg or png : https://web.dev/uses-webp-images/ )~
Moved to: https://github.com/ooni/ooni.org/issues/390

~Set `alt = "Image description"` for images: `![Image-description](/brief-image-description.png)`~
Moved to: https://github.com/ooni/ooni.org/issues/390

- [x] Add `sitemap.xml` : https://search.google.com/search-console/sitemaps?resource_id=sc-domain%3Aooni.org&hl=en : https://support.google.com/webmasters/answer/7451001?hl=en

- [x] Add `robots.txt` : https://support.google.com/webmasters/answer/6062608?hl=en

- [x] Redirect to the new address via `HTTP 301` . Two addresses with the same content are the worst. But OONI has three (or more) addresses ( `ooni.torproject.org` - `ooni.io` - `ooni.org` )  . Only one address must be used. 

**Medium priority:**

~On all pages, the main title should be in the `<h1>` tag.~
Moved to: https://github.com/ooni/ooni.org/issues/391

- [x] Avoid having multiple css files as much as possible. (Also js files) : https://gohugo.io/hugo-pipes/minification/ : https://web.dev/critical-request-chains/

~More favicon should be added: https://realfavicongenerator.net/favicon_checker?protocol=http&site=ooni.org#.Xfdm5FVfjIc~
Moved to: https://github.com/ooni/ooni.org/issues/392

- [x] Display image in link preview: https://discourse.gohugo.io/t/display-first-image-in-post-in-link-preview/10541/6

[...]

Issue #291, opened 2019-10-16
https://www.ooni.io is giving certificate errors


p1g30n:
Visiting https://www.ooni.io gives me `SSL_ERROR_BAD_CERT_DOMAIN` with a certificate that seems only valid for netlify.com and its subdomains.

xhdix:
Maybe @hellais should delete and re-add the www alias in Netlify.
And there is some other tips :
https://docs.netlify.com/domains-https/troubleshooting-tips

hellais:
It turns out that we didn't have the www in the certificate at all. I have re-added it.

xhdix:
@hellais the https://github.com/ooni/ooni.org/commit/9c0edd4ca7e7458957cda71ea07a8da6b0675a1c wasn't the best idea.
As @p1g30n  said https://github.com/ooni/ooni.org/pull/315#issuecomment-566756808 we must redirect **to** the www. (This is the first thing Google wants)
Also the answer of `dig ooni.org` still does not equal to `dig www.ooni.org`.

p1g30n:
@xhdix I wasn't sure about the best redirect solution when I made that comment, I just saw that https://www.ooni.io didn't redirect to https://ooni.io , which also meant it wouldn't finally redirect to https://ooni.org.
Do you know wether Google prefers less 301s (as with the current solution) or more accurate mapping of domains (e.g. https://www.ooni.io -> https://www.ooni.org -> https://ooni.org) ?

A third option would be that all domains first undo their www. subdomain and then redirect to https://ooni.org (as in https://www.ooni.io -> https://ooni.io -> https://ooni.org)

I'm unclear what the best option would be

hellais:
Hum, does this mean we must make the canonical version of our website be `https://www.ooni.org`?

I have never been a big fan of the `www` in front of domain names for websites, but mostly just for aesthetics. If making the canonical version of our site be `www.ooni.org` boosts our SEO, we should do it.

> Also the answer of dig ooni.org still does not equal to dig www.ooni.org.

The reason for this is that we can't put a CNAME on the apex of the domain, so we have to use the IP of one of the netlify hosts.

Netlify also suggest using www as the canonical version of your website, see: https://www.netlify.com/blog/2017/02/28/to-www-or-not-www/.

What do you folks think, should we `www` or not `www`?

xhdix:
@hellais If you check Google Webmaster Tools, all requests should be redirected to  `https://www.ooni.org`

[...]

PR #315, opened 2019-12-17
Redirect all addresses to ooni.org

xhdix:
part of #314


hellais:
🙏

p1g30n:
Maybe it makes since to add a redirect for https://www.ooni.io/ since it doesn't redirect to https://ooni.io/ . Doing that would probably require taking care of #291

xhdix:
@p1g30n  You're absolutely right.

PR #368, opened 2020-02-05
Bootstrap grid

hellais:
This is an experiment to investigate: https://github.com/ooni/ooni.org/issues/343.

It appears that it's not too much hard to just replace using the bootstrap grid and the changes required to our site are pretty small.

Still a few things need to be checked and fixed, but it doesn't seem too hard.


agrabeli:
Thank you @hellais , LGTM.

PR #369, opened 2020-02-06
Separate blog posts from research reports

hellais:
This adds a new menu item to the navbar for separating the research reports from the blog posts.


agrabeli:
@hellais thanks, LGTM.

PR #371, opened 2020-02-06
bootstrap fixes for mobile views

timepigeon:
The mobile views are currently broken, here are some fixes for the homepage (the sponsors could still use some love but this is all I got time for right now).
Generally the best way to work with bootstrap I've found is to go from the smallest device upwards - considering the largest view, bootstrap will automatically fall back to the next smaller one if none is defined, all the way down to the smallest one, if there are no definitions in between. 
So I usually start with the smallest, work my way upwards and only set other definitions when the (next) smallest one doesn't work on larger devices. Hope that makes any sense. Here's more info: https://getbootstrap.com/docs/4.0/layout/grid/

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -14,10 +14,10 @@
 
     <header>
       <div class="row">
-      <div class="col-3 mt-5">
-        <img class="logo" src="/images/ooni-header-mascot.png" srcset="/images/ooni-header-mascot-2x.png 2x" height="200" width="200"/>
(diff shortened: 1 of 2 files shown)


hellais:
Oh nice, thanks for this!

hellais:
If you have some spare cycles it would be great to get a review of: https://github.com/ooni/ooni.org/pull/370 & https://github.com/ooni/ooni.org/pull/369

PR #372, opened 2020-02-07
Make /post show the full index

hellais:
Set description and title to unique values


content/reports/_index.md
```diff
@@ -0,0 +1,4 @@
+---
+title: Research Reports
+description: OONI Research reports investigating internet censorship worldwide
```
hellais: @agrabeli you should review the text for these descriptions

PR #375, opened 2020-02-07
bootstrap adjustments for /data on mobile

timepigeon:
Adjustments for mobile views.

Added a helper class `.xs-center` for centering things on mobile (< 576px). (mostly to replace `class="d-block ml-auto mr-auto"` definitions, which are a bit verbose and can also interact in other views)


hellais:
Yes this looks very good. Thanks!

PR #387, opened 2020-02-13
Better ux

timepigeon:
Footer menu horizontal on sm and upwards, header menu horizontal on md and upwards . Added `white-space: nowrap;` to prevent line breaks in header menu items.


hellais:
Thanks for this!
I fixed the path for the fpu-logo (I think you had originally added a cropped version of it, but I could not find it in the history).

I have merged this into master.

PR #401, opened 2020-02-18
css adjustments for post tables

timepigeon:


timepigeon:
Sure. Is the padding more important in one direction? Maybe we could save some space in one dimension

timepigeon:
Assuming the horizontal padding is more important than the vertical one I went with 20px and 10px respectively for now, let me know if that works

hellais:
This looks great! Thanks for putting it together! 🐳

PR #404, opened 2020-02-24
Header fixes

timepigeon:
I'm getting console errors for highlight.js when loaded deferred, it resolves when loaded synchronously, but also seems to works asynchronously with the async attribute. I left it on the latter for now.
I took the opportunity to add corresponding mime types for stylesheets and scripts as well.


hellais:
Yes this is a good call.
Merging.

PR #418, opened 2020-03-06
Better images

hellais:
In this PR I look into implementing https://github.com/ooni/ooni.org/issues/390.

The first step in that direction is adding an end to end smoke test to ensure that no images break in moving to the new pattern.


hellais:
I am going to merge this and defer future work to #447

Issue #429, opened 2020-03-10
Update the install page with up to date instructions


hellais:
Currently the desktop install page (as several people have already pointed out since a while: https://github.com/ooni/ooni.org/issues/306) is a bit of a mess.

I did a mockup for how we could improve it some time ago:

![InstallPageRevamp](https://user-images.githubusercontent.com/424620/76301193-2575b200-62be-11ea-9fe6-b62c927d8fd8.png)

We should also consider just fixing up the obvious errors in it in the immediate short term even prior to implement the nice mockup mentioned above.

hellais:
Below are updated mockups for the various pages we need.

The idea is that users when they visit https://ooni.org/install will land on:
![Install Landing](https://user-images.githubusercontent.com/424620/76893214-45900d00-688c-11ea-9fa0-1cb8225fdb1c.png)

When they choose Desktop or Mobile they will land on the relevant page which has a phone or a desktop.
![Install Mobile](https://user-images.githubusercontent.com/424620/76893108-14174180-688c-11ea-988b-a84c9bacf799.png)

![Install Desktop](https://user-images.githubusercontent.com/424620/76893125-1aa5b900-688c-11ea-8d9e-a8fd9e98dac2.png)

When they are on the Desktop page and they click other platforms, then the modal will appear.

![Install Other Platforms](https://user-images.githubusercontent.com/424620/76893120-19748c00-688c-11ea-96e0-3446045fbfa7.png)

If they choose CLI they will end up on https://github.com/ooni/probe-cli.

Feedback from folks like @holantonela or others is very welcome :)

The copy is just a placeholder, it still needs some work on it from @agrabeli.

holantonela:
hey, it looks great! do you have it live somewhere so I can make in-context comments?

hellais:
Yes it should be possible to that here:
https://sketch.cloud/s/JOAde/a/0EV7x0
https://sketch.cloud/s/JOAde/a/l0ZGrO
https://sketch.cloud/s/JOAde/a/1DYPLy
https://sketch.cloud/s/JOAde/a/pReKAp

holantonela:
Thank you! just general comments were allowed, so I copy and pasted it in Marvel.

https://marvelapp.com/22374246/screen/67352108
https://marvelapp.com/22374246/screen/67352143

I think you need some titles [here](
https://user-images.githubusercontent.com/424620/76893120-19748c00-688c-11ea-96e0-3446045fbfa7.png
[...]

PR #446, opened 2020-03-22
New install page UI

SaptakS:
Refs #429 

Things to be done:
- [x] check if responsive and improve responsiveness
- [x] create the modal in desktop
- [x] Add icons for globe, fast and chat (cc: @hellais )

Following are the screenshots of how it looks in firefox inside debian

First Landing page:
Screenshot_2020-03-22 Install

Desktop landing page
Screenshot_2020-03-22 Measure internet censorship(3)

Mobile landing page
Screenshot_2020-03-22 Measure internet censorship(2)

Used `.md` files to just assign title, description, screenshot link and the buttons and links. Since the rest of the page is basically the same so put that inside the layout html.


hellais:
This looks great, thanks for putting it together!

I will take note of some things too look into:

* The background flourish should be fixed to the center of the phone or desktop screenshot. If I adjust the resolution of the window it goes out of alignment, see screenshot below:

Screenshot 2020-03-25 at 16 40 35

* The 2x version of the app store icons is actually not double the resolution, leading to downsampling on double density displays (ie. retina screens)

* An on-hover effect should be added to the primary action button to perform the download of the app. It can be something as simple as just adjusting the color to be little bit darker. Try with #056aa6 (or another of our darker blue choices in our palette: https://design.ooni.io/colors/).

* The contrast of the command line interface section is not great. This is more of a TODO for me to figure out how we can improve the readability of that section.

SaptakS:
Updated most of the designs and added the all-platform page. Need to still work on responsiveness (across all resizing of windows) and hiding buttons using JS in desktop installation page.

hellais:
It's starting to take a good shape!

I pushed to your branch some small fixes for the background flourish and the phone mockup image.

Having looked at the code a bit more, I think it might be a bit easier to reason about it if we split out the desktop and mobile specific code inside of partials rather than having those many if else blocks.

Perhaps we could have the following partials:
- install-mobile.html
- install-desktop.html
- install-all.html

There will be a bit more code duplication, but I think we gain in terms of readability.

What do you think?

SaptakS:
Agreed. I was just trying to reduce the duplications. But I guess using partials makes more sense and readable. I will update accordingly.

SaptakS:
@hellais I have added some `overflow-x: hidden` to prevent the overflow caused by the flourish positioning. Also, made it position based on `right` rather than `left` to keep it at same position in both mobile and desktop, irrespective of the grid layout.

I have also moved the 3 conditional stuffs, into 3 different partials and included the partials in single.html based on the conditions.

hellais:
Thanks for putting this together @SaptakS it's looking great!

I did a pass of reviewing your branch and made some small tweaks to the responsiveness of the pages and fixed a few issues with the images appearing oddly scaled on some resolutions.

I also tweaked a bit the JS code you wrote for the OS detection to not rely on jQuery or similar.

Here are a few things which I think we could address, if we have time, but I don't think are blocking on shipping this:
* [ ] The `feature-container` (Check which websites are blocked, how fast is your network, etc.) should be single column when scaled down to very small screens (sm and lower)
* [ ] The "download OONI Probe for mac" button on the desktop install page should be horizontally centered when the desktop screenshot disappears on smaller viewports
* [ ] I should review the font selection on these page. We are currently mixing a serif with a san-serif font. We should only really be using the serif font for articles. This might actually be done as a site-wide change.
* [ ] It would be great if all the text content were inside of a markdown file for easy editing and localisation support (ex. the stuff currently in install/single.html)

I would say that the only thing blocking on shipping this is getting the final copy from @agrabeli and obviously cutting an official 3.0 release of OONI Probe Desktop.

We can potentially even roll this out earlier and just disable the OONI Probe Desktop link from the install page.

SaptakS:
> I also tweaked a bit the JS code you wrote for the OS detection to not rely on jQuery or similar.

@hellais makes sense. I saw that jQuery was available so I used it anyways, else I do support the use of vanilla JS over relying on any other library or framework myself.

SaptakS:
@hellais following have been done:
- [x] The feature-container (Check which websites are blocked, how fast is your network, etc.) should be single column when scaled down to very small screens (sm and lower)
- [x] The "download OONI Probe for mac" button on the desktop install page should be horizontally centered when the desktop screenshot disappears on smaller viewports

Also made the command line screenshot hidden in `xs` given that it wasn't being visible anyways.

agrabeli:
@SaptakS @hellais  Thank you for the excellent work!

I have written the copy for these pages here: https://docs.google.com/document/d/1xcn29T-SywfLI5n3vonCSV7N0q1edTTL0jOpPUM-Ps8/edit?usp=sharing

agrabeli:
Thank you @SaptakS for updating the new Install pages with the copy!

I have some feedback:

1) Landing Install page: Please remove the full stop at the end of "Contribute to the world's largest open dataset on internet censorship". 

install

2) Mobile Install page: The end of the subtitle ("Contribute to the world's largest open dataset on internet censorship") is buried under the rainbow, see:

mobile-rainbow

3) Tests banner ("Check which apps are blocked"): Please replace "run OONI Probe" with "Run OONI Probe".

apps

4) "Share evidence of internet censorship with the world" section: Please delete the word "the" before "transparency" so that it reads as follows: "By running OONI Probe, you help increase transparency of internet censorship."

explorer-install

5) `Explore OONI measurements from around the world >>` => Please link to https://explorer.ooni.org/ (rather than to the Mobile Install page)

6) Desktop Install page: The end of the subtitle ("Contribute to the world's largest open dataset on internet censorship") is buried under the rainbow, see:

desktop-rainbow

7) Desktop Install page: Replace "other platforms" with "Other platforms".

platforms

8) Desktop: `Explore OONI measurements from around the world >>` => Please link to https://explorer.ooni.org/ (rather than to the Desktop Install page)

9) All platforms page: `Explore OONI measurements from around the world >>` => Please link to https://explorer.ooni.org/ (rather than to the Install All page)

Thank you. :smiley_cat:

agrabeli:
Thank you @SaptakS , it looks perfect now!

Some other minor feedback (cc' @hellais ):

1. Update the OONI Probe desktop app screenshots so that they include the latest version (which, for example, has the OONI logo on it)

2. Include a better quality screenshot/image of OONI Explorer in the header below ("Share evidence with the world" section). Perhaps consider including the 1st image from here: https://ooni.org/post/next-generation-ooni-explorer/

3. Consider including the rainbow in the "All Platforms" page (as it looks quite empty in comparison to the other Install pages).

hellais:
> Update the OONI Probe desktop app screenshots so that they include the latest version (which, for example, has the OONI logo on it)

Yes this is a good point I will update it prior to merging.

> Include a better quality screenshot/image of OONI Explorer in the header below ("Share evidence with the world" section). Perhaps consider including the 1st image from here: https://ooni.org/post/next-generation-ooni-explorer/

I am not sure what you mean by this exactly. Is the issue with the resolution of the screenshot, the content of the page or the layout?
The image you link to is on a blue background and in a very different format, so it will not work on that page.

I would say unless it's a technical issue (ex. the resolution being bad) that can be easily fixed with a drop-in replacement, we should postpone this to future versions.

> Consider including the rainbow in the "All Platforms" page (as it looks quite empty in comparison to the other Install pages).

This also should be deferred to future work as it's more work.

PR #451, opened 2020-03-27
New Tor & Psiphon test descriptions

agrabeli:
I have written descriptions for the new Tor & Psiphon tests.

### DIFF
--- a/content/nettest/psiphon.md
+++ b/content/nettest/psiphon.md
@@ -23,11 +23,11 @@ can run to measure the reachability of the Psiphon app.
 The **OONI Probe Psiphon test** provides an automated way of examining whether the
 [Psiphon app](https://psiphon.ca/) works in a tested network.
 
-This [test](https://github.com/ooni/probe-engine) starts off by compiling the Psiphon codebase in an attempt to
-bootstrap Psiphon. Once the test has bootstrapped and created the Psiphon
-tunnel, it then checks whether it actually works. To determine this, the test
-attempts to fetch a webpage to see if it’s possible to use Psiphon in order to
-connect to websites on the internet. 
+This [test](https://github.com/ooni/probe-engine) starts off by using Psiphon
+code to bootstrap a Psiphon tunnel. Once the test has bootstrapped and created
+the Psiphon tunnel, it then checks whether it actually works. To determine this,
+the test attempts to fetch a webpage to see if it’s possible to use Psiphon in
+order to connect to websites on the internet. 
 
 As part of this process, there are 3 possible outcomes:
 
--- a/content/nettest/psiphon.md
+++ b/content/nettest/psiphon.md
@@ -7,14 +7,40 @@ groups: ["proxies"]
 
 # Psiphon
 
-This test provides an automated way of examining whether Psiphon works in a
-tested network.
+## About Psiphon
 
-[Psiphon](https://psiphon.ca/) is a free and open source tool that utilises SSH,
-VPN and HTTP proxy technology for censorship circumvention.
+[Psiphon](https://psiphon.ca/) is a free and open source tool that utilizes VPN, SSH, and HTTP proxy
+technology for censorship circumvention. The [Psiphon VPN](https://psiphon.ca/en/download.html) essentially serves as a
+tunnel that enables you to circumvent censorship because you access resources on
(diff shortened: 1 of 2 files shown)


agrabeli:
@bassosimone Is it accurate to say that if *any* of the steps in the Tor test fail, then Tor doesn't work in the tested network (as described towards the end of the Tor test description)?

content/nettest/psiphon.md
```diff
@@ -7,14 +7,40 @@ groups: ["proxies"]
...
+In an attempt to limit censorship circumvention, some governments have [blocked](https://ooni.org/post/iran-internet-censorship/#anonymity-and-circumvention-tools)
+access to Psiphon. OONI has therefore developed an [OONI Probe](https://ooni.org/install/) test that anyone
+can run to measure the reachability of the Psiphon app.
+
+## Psiphon test
+
+The **OONI Probe Psiphon test** provides an automated way of examining whether the
+[Psiphon app](https://psiphon.ca/) works in a tested network.
+
+This [test](https://github.com/ooni/probe-engine) starts off by compiling the Psiphon codebase in an attempt to
```
bassosimone: This [test](https://github.com/ooni/probe-engine) starts off by using Psiphon code to bootstrap a Psiphon tunnel.

content/nettest/tor.md
```diff
@@ -0,0 +1,63 @@
...
+OONI has developed an [OONI Probe](https://ooni.org/install/) test that anyone can run to measure the
+blocking of [Tor](https://www.torproject.org/).
+
+## Tor test
+
+The **OONI Probe Tor test** provides an automated way of examining whether [Tor](https://www.torproject.org/)
+works in a tested network.
+
+To this end, this [test](https://github.com/ooni/probe-engine) examines the reachability of a set of services that allow
+Tor to work. These include the following protocols:
```
bassosimone: protocols -> services

content/nettest/tor.md
```diff
@@ -0,0 +1,63 @@
...
+* obfs4 (Tor bridge that speaks the OBFS4 protocol)
+
+By measuring the reachability of selected Tor directory authorities and bridges,
+this test evaluates whether they can be used and, by extension, whether Tor
+works within the tested network.
+
+More specifically, this test attempts to perform the following actions from the
+vantage point of the user:
+
+* HTTP GET request to the ``tor/status-vote/current/consensus.z`` resource, which
```
bassosimone: replace double backticks with single backticks?

content/nettest/tor.md
```diff
@@ -0,0 +1,63 @@
...
+
+* OR port (used by Tor bridges) 
+
+* OR port of directory authorities (used by Tor clients) 
+
+* obfs4 (Tor bridge that speaks the OBFS4 protocol)
+
+By measuring the reachability of selected Tor directory authorities and bridges,
+this test evaluates whether they can be used and, by extension, whether Tor
+works within the tested network.
```
bassosimone: I don't think this test allows us to conclude whether Tor works. If it fails we know it would not work because some resources are blocked, and this would inform circumvention policies. If it works, it may instead be that there are other sources of blocking that we're not checking for. This is why this test will periodically need to be extended to include more checks that allow to characterise more forms of blocking. (That is, we're not measuring whether Tor does not work but whether there are specific policies designed to prevent Tor from working.)

content/nettest/psiphon.md
agrabeli: Thanks, I've edited this.

content/nettest/tor.md
agrabeli: Thanks.

content/nettest/tor.md
agrabeli: Thanks.

content/nettest/tor.md
agrabeli: Thanks for the clarification.

agrabeli:
Thank you @bassosimone for the review! I have made edits based on your feedback.

I have also edited the Tor test description to say that Tor may not work if it is being blocked in ways that aren't being measured by this test (see: https://github.com/ooni/ooni.org/pull/451/commits/49f0201b59b65b66729d22da9f5e2bf8ec5218bb).

PR #468, opened 2020-04-08
OONI Probe Desktop Launch Announcement

agrabeli:
This will be **merged & published on 21st April 2020**, along with the public launch of the new OONI Probe desktop app.

Before we merge and publish this post, we need to:

- [x] Review (and potentially update) the listed translated languages

- [x] Embed 2 **Download buttons** for Windows and macOS (row 15)

- [ ] Replace screenshots with latest version of OONI Probe Desktop (e.g. including OONI logo) and better quality images, preferably some showing blocking (cc' @hellais )


hellais:
TODO
* [ ] For some reason the post is not appearing in the listing, we should fix that
* [ ] The post should be adjusted to the new pattern, see https://github.com/ooni/ooni.org/tree/master/content/post/covid-19-ooni-measurement as an example

agrabeli:
I have removed reference to the Arabic & Farsi translations because the desktop app does not include RTL support yet (this will be added after the public launch).

hellais:
This is rebase and merged. Closing.

PR #476, opened 2020-04-14
Policy for URL prioritization

agrabeli:
This PR includes a blog post that describes OONI's upcoming new smart URL list system, and links to OONI's policy for URL prioritization.


agrabeli:
@hellais the netlify deployment doesn't seem to work here either. :/

PR #478, opened 2020-04-15
Updated the OONI Data Policy

agrabeli:
This PR includes updates to the OONI Data Policy. This is now version 1.3.0.


agrabeli:
@hellais I'd appreciate your careful review of the updated OONI Data Policy. 

Some things to look our for include:

* References to **push notification** and relevant sections under "Data We Collect", "Data We Store", "Data We Publish"

* Analytics section => Is everything there accurate? Anything else worth mentioning?

* "Network information" section (under "Data We Collect") => Is everything there accurate?

* Section on "crash reports" => Is everything accurate? Anything missing?

Thank you!

content/about/data-policy.md
```diff
@@ -4,123 +4,254 @@ aliases: ["/data-policy"]
...
+personally-identifiable information. 
+
+### 3. User Control
+
+Running OONI Probe can potentially be [risky](https://ooni.org/about/risks/),
+and different users have different threat models. It's therefore important to us
+that OONI Probe users are able to customize their testing depending on their
+threat model and what they feel comfortable with. 
+
+To this end, we provide settings in the OONI Probe apps that enable users to **opt
```
hellais, line 77: Can we replace in the data principles section all reference to "OONI Probe" with "OONI software". I think these principles apply to any OONI software product, not just OONI Probe.

content/about/data-policy.md
```diff
@@ -4,123 +4,254 @@ aliases: ["/data-policy"]
...
+### 5. Accessible and Usable
+
+We openly publish all OONI measurements in machine-readable format, so that they
+can more easily be accessed, used, and integrated into other systems.
+
+To enable researchers and data scientists to make use of OONI data, we have
+created an [OONI API](https://api.ooni.io/) for downloading the raw measurements in JSON format. For
+batch consumption of OONI data, we have created a system that enables third
+parties to set-up their own instance of the [OONI PostgreSQL Metadb](https://github.com/ooni/sysadmin/blob/master/docs/metadb-sharing.md). The whole
+OONI dataset can also be fetched from the [ooni-data Amazon S3 bucket](https://ooni.org/post/mining-ooni-data).
```
hellais, line 96: Maybe instead of referencing specific ways to use the OONI data which may change in the future, we can instead mention that we offer many different options for accessing OONI data including in batch. Then just limit to link to the ooni.org/data/ page.

content/about/data-policy.md
```diff
@@ -4,123 +4,254 @@ aliases: ["/data-policy"]
...
+By default, we collect your corresponding network **[Autonomous System Number (ASN)](https://ooni.org/support/glossary/#asn)**
+when you run OONI Probe. 
+
+Collecting ASNs helps our work by revealing the identity
+of network providers that engage in censorship or other forms of network
+interference. For example, if you run OONI Probe on Vodafone Italia, we would collect
+"AS30722", which is the corresponding ASN.
+
+We also collect information to identify if an OONI Probe user is on a **WiFi network** or
+using **mobile data**.
```
hellais, line 184: I'm not sure if it's possible to opt out of this. I documented that here: https://github.com/ooni/probe/issues/1112.

If it turns out that is not the case, then we should clarify in this section that you can only opt out of collecting the AS number, but not if the measurement was done on mobile or wifi.

content/about/data-policy.md
```diff
@@ -4,123 +4,254 @@ aliases: ["/data-policy"]
...
-ooniprobe is testing employ tracking technologies or contain custom content.
+or other metadata of OONI measurements. This data, for example, might be collected if
+the [websites](https://ooni.org/support/faq/#which-websites-will-i-test-for-censorship-with-ooni-probe) OONI Probe is testing employ tracking technologies or contain custom content.
 
 By default, we do *not* collect your IP addresses. You can, however, opt-in to
 provide your individual IP address (to enhance the accuracy of information) by
-[editing the ooniprobe  configuration](https://github.com/TheTorProject/ooni-probe#configuring-ooniprobe) file inside of `~/.ooni/ooniprobe.conf`.
-Alternatively, you can opt-in to send us this information through the setup
-settings of OONI's web user interface or OONI's mobile app.
+enabling this option in the settings of the OONI Probe apps.
```
hellais, line 233: We should put a sentence here about the fact that the performance tests will result in the users IP address always being collected by MLab, though we are discussing with them to try to change that.

content/about/data-policy.md
```diff
@@ -4,123 +4,254 @@ aliases: ["/data-policy"]
...
+collect information required for identifying and authenticating probes to send out push notifications. 
+
+This may include:
+
+* Information about the OONI Probe installation (such as software version,
+supported tests, platform, device language, device timezone);
+
+* Network information (such as probe ASN and network type) and country code;
+
+* Push notification token.
```
hellais, line 247: This is good.

content/about/data-policy.md
```diff
@@ -140,44 +271,51 @@ try to make sense of the data and to find answers to the following questions:
...
-to allow third parties to conduct independent studies, verify our findings, and
-answer other research questions.
+We publish ALL of the OONI Probe network measurement data that we have collected
+and stored to allow third parties to conduct independent studies, verify our
+findings, and answer other research questions.
 
-Currently, we publish all measurement data both through our [OONI API](https://api.ooni.io/) and through an interactive
-visualization called [OONI Explorer](https://explorer.ooni.org/world/).
+Currently, we publish all OONI measurement data both through our [OONI API](https://api.ooni.io/) and through an interactive
+web interface, called [OONI Explorer](https://explorer.ooni.org/).
```
hellais, line 297: We should also link to the ooni.org/data page that includes also reference to s3.

content/about/data-policy.md
agrabeli: Good point! I've edited accordingly.

content/about/data-policy.md
agrabeli: Addressed.

content/about/data-policy.md
agrabeli: Thanks for the clarification.

content/about/data-policy.md
agrabeli: That's a good point! I've edited accordingly.

content/about/data-policy.md
agrabeli: Done.

agrabeli:
@hellais can you please include the Matomo analytics opt-out box at the end of the Data Policy page?

PR #485, opened 2020-04-24
Fix padding of footer

hellais:
Fixes: ooni/ooni.org#469


hellais:
I am going to self merge this.

PR #503, opened 2020-05-21
Report on social media blocking in Burundi

agrabeli:


agrabeli:
@hellais How can we make the chart bigger so that it's possible to read the information included in it?

Issue #480, opened 2020-04-17
Polish the website navbar


hellais:
The navbar of the website could use with some polishing.

This issue is to do that.

Things to look at are:
* [ ] Menu items on top need to be centered too (suggested by @agrabeli )

hellais:
This is done

PR #518, opened 2020-06-08
center nav menu for all views

timepigeon:
Fix for #480


hellais:
I tested it and it looks good to me! Merging!

PR #555, opened 2020-07-02
Updated OONI Data Policy

agrabeli:
Following the change of MaxMind's license, the OONI team created our own ASN database.

I have updated the OONI Data Policy to reflect this change.

### DIFF
--- a/content/about/data-policy.md
+++ b/content/about/data-policy.md
@@ -128,8 +128,6 @@ by **unchecking the opt-out box** at the end of this Data Policy.
 You can opt out of our use of analytics on the OONI Probe mobile and desktop
 apps by disabling the collection of app usage statistics in the app settings. 
 
-We will never link OONI Probe app usage metrics to your identity.
-
 We will notify you of any future changes to our use of analytics through an
 update to this Data Policy.
 
--- a/content/about/data-policy.md
+++ b/content/about/data-policy.md
@@ -4,7 +4,7 @@ description: This Data Policy discloses and explains what data the OONI project 
 aliases: ["/data-policy"]
 ---
 
-**Last modified:** July 3, 2020
+**Last modified:** July 15, 2020
 
 **Version:** 1.3.2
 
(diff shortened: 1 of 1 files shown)


agrabeli:
@bassosimone @hellais The latest edits to the OONI Data Policy say that we're using our own ASN database, and that you cannot opt-out of submitting country code information. Please review the changes carefully before we publish. :)

agrabeli:
@hellais With my latest commit, I have updated the OONI Data Policy to discuss our use of Countly and collection of app usage statistics. Is everything framed accurately and does it cover everything? Or are there things that I've missed?

content/about/data-policy.md
```diff
@@ -111,23 +111,34 @@ collected from around the world.
...
 Through the use of analytics, our goal is to improve our understanding of user
 needs, improve OONI tools, and better serve the OONI community.
 
 We do not record the full IP address of users (which is "anonymised" to the
 first 3 octets, ex. 123.45.67.0).
 
+We use [Matomo analytics](https://matomo.org/) -- a [GDPR compliant, free and
+open source tool](https://matomo.org/why-matomo/) -- to collect some information
+about visitors to [OONI Explorer](https://explorer.ooni.org/) and our [website](https://ooni.org/), as well as to collect [OONI Probe desktop
+app](https://ooni.org/install/desktop) usage statistics.
```
hellais, line 123: Can we make this a bit more high level and just speak about a self hosted analytics platform?

content/about/data-policy.md
```diff
@@ -111,23 +111,34 @@ collected from around the world.
...
 by **unchecking the opt-out box** at the end of this Data Policy.
 
-You can opt out of our use of analytics on the [OONI Probe desktop app](https://ooni.org/install/desktop) by disabling the collection of app usage
+You can opt out of our use of analytics on the [OONI Probe desktop
+app](https://ooni.org/install/desktop) by disabling the collection of app usage
 statistics in the app settings. 
 
+To better understand how the [OONI Probe mobile
+app](https://ooni.org/install/mobile) is used, we collect app usage statistics
+through the use of [Countly analytics](https://count.ly/).
```
hellais, line 134: Again, here the key point is that it's self hosted. It's especially important in the case of countly because they also offer a hosted solution, which is not what we are using.

content/about/data-policy.md
```diff
@@ -215,6 +229,18 @@ about why and how a specific OONI Probe function failed to work as expected.
...
 advanced settings of the OONI Probe apps.
 
+#### App usage statistics
+
+By default, we collect OONI Probe app usage statistics in order to better
+understand how our apps are used, improve our understanding of user needs,
+improve OONI Probe apps, and better serve the OONI community.
+
+On the [OONI Probe desktop app](https://ooni.org/install/desktop), we use [Matomo analytics](https://matomo.org/), while on the [OONI Probe
+mobile app](https://ooni.org/install/mobile), we use [Countly analytics](https://count.ly/).
```
hellais, line 239: Replace wording of countly with "self hosted analytics platform"

content/about/data-policy.md
agrabeli: Sure. I have edited this accordingly.

content/about/data-policy.md
agrabeli: Thanks, I have edited this.

content/about/data-policy.md
agrabeli: Thanks, I have edited this.

hellais:
I think we might also need to add something about the data we collect in relation to push notifications
For example that when a user enables push notifications, in order to send push notification we collect their approximate geographical location (based on their IP) so that we can send geo targetted push notifications
We also should specificy somewhere what our data retention policy is, as in after how long we will delete the data we collect
In relation to app analytics and push notifications we might also need to offer the ability for users to request immediate deletion of all data about them.

@lorenzoPrimi  do you know how we can delete all user data? Is there some user ID thing that countly generates that user need to provide us with in order to zap all the data we have about them?

agrabeli:
Thanks @hellais for your feedback. In my latest commit (see: https://github.com/ooni/ooni.org/pull/555/commits/efd1a39801f002bea1507ce9df3384488262322e) I updated the push notifications section of the Data Policy to mention that we may collect the approximate geographical location (based on your comment).

We have a section on data retention ("Data we Store"), and that has been edited to mention data collected from push notifications as well. However, it's unclear to me how to edit further. How long would the data be retained for? When would it be deleted? Further details would help be edit further, thanks. 

^^ cc' @lorenzoPrimi

content/about/data-policy.md
```diff
@@ -4,9 +4,9 @@ description: This Data Policy discloses and explains what data the OONI project
 aliases: ["/data-policy"]
 ---
 
-**Last modified:** April 22, 2020
+**Last modified:** July 22, 2020
 
-**Version:** 1.3.1
+**Version:** 1.3.2
```
hellais, line 9: Do we maybe want to bump the version number to 1.4.1? I think it's a big enough change to warrant a minor version bump and not just a patch level bump.

agrabeli: That's a good point. I just edited it to 1.4.0.

PR #667, opened 2020-11-02
Replace mockups of phones

hellais:
Fixes: https://github.com/ooni/ooni.org/issues/511


agrabeli (approved):
@hellais Looks great! Thanks so much for working on this. :+1:

PR #705, opened 2020-11-23
Update Data Policy based on OONI Probe Mobile 2.7.1 release

agrabeli:
We have released **OONI Probe Mobile 2.7.1** which includes the following noteworthy changes:

- We have **removed the "Include Network Info" setting (which enabled you to opt-out of ASN collection)**: Now ASN information is collected by default. We decided to do this because measurements that don't include ASN information are non-actionable.

- We have **removed the setting which enabled you to share your IP address** (on an opt-in basis): We decided to remove this setting because we think the potential risk associated with sharing IP addresses is greater than the benefit (over the last years, when users chose to share their IP addresses, we did not find them particularly useful).

Through these changes (in OONI Probe Mobile 2.7.1), we aim to make OONI Probe measurements both more secure and more actionable. 

These changes have been reflected in the OONI Data Policy, where I specify that the settings to opt-out of ASN collection and to opt-in to IP address collection are currently limited to the OONI Probe desktop app (since we have now removed them on mobile). 

In the next weeks, we plan to remove these options on desktop as well (at which point I will make further edits to the OONI Data Policy).

### DIFF
--- a/content/about/data-policy.md
+++ b/content/about/data-policy.md
@@ -4,9 +4,9 @@ description: This Data Policy discloses and explains what data the OONI project 
 aliases: ["/data-policy"]
 ---
 
-**Last modified:** July 22, 2020
+**Last modified:** November 24, 2020
 
-**Version:** 1.4.0
+**Version:** 1.4.1
 
 This Data Policy discloses and explains what data the [Open Observatory of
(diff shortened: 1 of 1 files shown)


bassosimone (approved):
🐳

PR #706, opened 2020-11-24
Bump probe-desktop download version to 3.1.1

sarathms:


sarathms:
Merging this because its a trivial change,

PR #831, opened 2021-03-29
Fix markdown links in post

anadahz:


anadahz:
The links are not being parsed correctly in: https://ooni.org/post/ooni-cryptorave-2017/
Fixed in 9880f96 .

hellais:
LGTM!

PR #863, opened 2021-05-04
chore(debian): post-bintray install instructions

bassosimone:
Part of https://github.com/ooni/probe/issues/1440


bassosimone:
@hellais I guess we also want to mention this migration, don't we?

hellais:
LGTM

PR #1043, opened 2021-12-07
Restructured and updated Nettests section

agrabeli:

### DIFF
--- a/content/nettest/web-connectivity.md
+++ b/content/nettest/web-connectivity.md
@@ -3,7 +3,7 @@ title: "Web Connectivity"
 short_description: "This test checks whether access to websites
 is blocked by means of DNS tampering,
 TCP/IP blocking or by a transparent HTTP proxy."
-groups: ["website"]
+groups: ["websites"]
 date: "0000-01-01"
 ---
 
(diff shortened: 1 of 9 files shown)


content/nettest/web-connectivity.md
```diff
@@ -3,7 +3,7 @@ title: "Web Connectivity"
 short_description: "This test checks whether access to websites
 is blocked by means of DNS tampering,
 TCP/IP blocking or by a transparent HTTP proxy."
-groups: ["website"]
+groups: ["websites"]
```
hellais, line 6: This needs to be `website`, if you change it to `websites` it will not match in the matcher below.

hellais: Ah I see you now changed the key below, so I guess it will work now.

agrabeli: @hellais I just edited it to `websites` in nettests.html so that it can be consistent with the id: https://github.com/ooni/ooni.org/pull/1043/commits/947600143d44bfeebb870f3cc4b71750d852bf9a

hellais: Yes I saw it now 👍

PR #1075, opened 2022-01-19
Report http3 decreased table width

kelmenhorst:
This is a low priority improvement for the new HTTP/3 measurement post.


agrabeli:
Thank you @kelmenhorst for changing the table size (much better)! <3

PR #1170, opened 2022-05-23
Add extrepo instructions

### LINKED ISSUE #1162: apt-key is deprecated on Debian Bullseye and Ubuntu Jammy
Following the instructions [here](https://ooni.org/install/cli/ubuntu-debian) results in:

```
~$ sudo apt-key adv --verbose --keyserver hkp://keyserver.ubuntu.com --recv-keys 'B5A08F01796E7F521861B449372D1FF271F2DD50'
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
<snip>
```
Processing the key with this command still works but on Ubuntu i now get an annoying warning everytime i run `sudo apt update`.

FedericoCeratto:
FIxes https://github.com/ooni/ooni.org/issues/1162

### DIFF
--- a/content/install/cli/ubuntu-debian.md
+++ b/content/install/cli/ubuntu-debian.md
@@ -4,26 +4,25 @@ description: How to install the OONI Probe CLI on Debian/Ubuntu Linux
 platform: cli
 ---
 
-1) Install the bintray GPG key:
+There are different ways to install OONI Probe CLI on Debian and derivatives.
+
+#### Using extrepo
+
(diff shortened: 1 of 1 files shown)


bassosimone (changes requested):
I think we should figure out why I get such an old version of ooniprobe-cli when using extrepo. I have documented the procedure I followed and the packages I've got.

McLP2:
The variant I proposed in #1580 (sorry, I did not look before if someone was already working on it) is probably the easier solution.

hellais:
Superseeded by https://github.com/ooni/ooni.org/pull/1580

PR #1199, opened 2022-07-28
Blog post and user guide for Test Lists Editor

agrabeli:

This PR includes a blog post for the announcement of the public launch of the Test Lists Editor. The post also includes a user guide.


agrabeli:
It seems there's some issue with the Netlify deployment, but when I reviewed this publication from a localhost and it seemed to be fine. I therefore assume that if I merge this PR, the post should be published OK.

PR #1213, opened 2022-08-15
make nonexisting image path break with informative message

ainghazal:
the error message for builds is quite cryptic, since it will complain that Fit cannot be found in a resource (because it's null if the GetMatch function fails).

this will leave a FIXME in the rendered output, which imho is more informative that the previous state.

edit: changed to log a fatal error.

### DIFF
--- a/layouts/shortcodes/img.html
+++ b/layouts/shortcodes/img.html
@@ -1,5 +1,9 @@
 {{ $img := $.Page.Resources.GetMatch (.Get "src")}}
 
+{{ if eq $img nil }}
+{{ errorf "Cannot load image (check spelling?): <%s%s>" ($.Page.RelPermalink) (.Get "src") }}
+{{ else }}
+
 {{ $img600 := $img.Fit "1000x800" }}
 {{ $img1200 := $img.Fit "2000x1600" }}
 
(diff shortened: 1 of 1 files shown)


layouts/shortcodes/img.html
```diff
@@ -1,5 +1,11 @@
 {{ $img := $.Page.Resources.GetMatch (.Get "src")}}
 
+{{ if eq $img nil }}
+<div>
+    <p><span style="color: red; font-weight:bold;">FIXME: </span>missing image path <span style="font-weight: bold">{{ (.Get "src") }}</span></p>
+</div>
+{{ else }}
```
bassosimone, line 7: Is it possible to make this crash/panic with an informative error message instead?

ainghazal: sure - I will change it to output a fatal error (that causes build to fail).

ainghazal:
```
WARN 2022/08/17 11:56:29 The "tweet" shortcode will soon require two named parameters: user and id. See "/home/kali/ooni/ooni.org/content/post/ethiopia-internet-shutdown-amidst-recent-protests.md:156:1"
ERROR 2022/08/17 11:56:29 Cannot load image (check spelling?): </support/ooni-probe-desktop/images/install-desktop-foobar.png>
Error: Error building site: logged 1 error(s)
```

bassosimone: Thank you!

hellais (approved):
This looks good to me. Thanks for putting it together!

PR #1217, opened 2022-08-29
doc(ooni-probe-cli): see automated results on debian

bassosimone:
This diff documents the (slightly hackish) command with which a
ooniprobe user could see automated-tests results on debian.

We look forward to make this easier, but there's no reason to
avoid documenting the working command to see results.

Thanks to the users who prodded us to document this functionality!

Part of https://github.com/ooni/probe/issues/1771.

(While there, zap empty spaces at end of line, which vscode did
for me, and which seems nice to do.)


sloncocs (approved):
Thank you Simone!

PR #1239, opened 2022-09-16
Updated settings for the Automated testing [OONI Probe Mobile user guide]

sloncocs:


agrabeli (approved):
@sloncocs excellent, thanks so much! <3

PR #1240, opened 2022-09-16
Updated settings for the Automated testing [OONI Probe Desktop user guide]

sloncocs:


agrabeli:
@sloncocs perfect, thanks a million for these updates! <3

PR #1249, opened 2022-09-28
fix: correct obvious typo

bassosimone:


sloncocs:
Thank you for noticing and correcting this! <3

PR #1265, opened 2022-10-19
Mobile user guide Farsi localisation [INTERIM SOLUTION]

sloncocs:


agrabeli (approved):
@sloncocs huge thanks to you and the translator! <3

PR #1269, opened 2022-10-25
Fix text alignment for rtl languages

majakomel:
Ticket: https://github.com/ooni/ooni.org/issues/1264


sloncocs:
Thank you @majakomel! ❤️

PR #1270, opened 2022-10-25
Optimize css requests

akronb:
This PR adds bundling and minification for base CSS.


akronb:
Hm, I've mounted `static` directory as `assets` and it worked on local, but crashed on Netlify. To fix that, I moved `css` back to `assets`.

I also saw that `zeptojs` is loaded, but I do not understand whether it is used. Maybe it can be removed?

hellais:

> Hm, I've mounted static directory as assets and it worked on local, but crashed on Netlify. To fix that, I moved css back to assets.

This might have something to do with the fact we are using an older version of hugo compared to what you might be using locally. Specifically we are still on `0.76.5` (https://github.com/ooni/ooni.org/blob/master/netlify.toml#L6). We should eventually upgrade to a newer version, but I fear several things will break in the process.

> I also saw that zeptojs is loaded, but I do not understand whether it is used. Maybe it can be removed?

Yes, you are right. This dependency is not used anymore in the site (IIRC it was used in the past for the install pages, but has since been removed).

I think we can safely drop it from inclusion.

I took a look at the diff and did some light testing of it. To me this looks good to merge.

If you want I can wait for the removal of zepto in this PR or we can land this and do that as a separate one.

akronb:
@hellais I removed `zeptojs`. If everything is okay, you can merge it.

content/support/ooni-probe-desktop/index.md
```diff
@@ -2,11 +2,11 @@
 title: "User Guide: OONI Probe Desktop App"
 description: "This user guide provides step-by-step instructions on how to use the OONI Probe desktop app."
 ---
-**Last updated:** 18th October 2022
+**Last updated:** 25th October 2022
 
 **OONI Probe Desktop version:** 3.7.0
 
-**Access the OONI Probe Desktop user guide in other languages:** [فارسی](https://github.com/ooni/ooni.org/blob/master/content/support/ooni-probe-desktop/index.fa.md)
+**Access the OONI Probe Desktop user guide in other languages:** [فارسی](https://ooni.org/fa/support/ooni-probe-desktop)
```
hellais, line 9:
```suggestion
**Access the OONI Probe Desktop user guide in other languages:** [فارسی](/fa/support/ooni-probe-desktop)
```

hellais:
LGTM!

Thanks for working on this!

PR #1284, opened 2022-11-13
OONI Probe Desktop user guide (RU)

sloncocs:


agrabeli (commented):
Thanks @sloncocs for the Russian translation! <3

My only feedback is that I'd suggest keeping the previous update date, as that pertains to the last update to the original (English) version of the user guide.

It's unclear to me though why we're unable to preview the Russian translation, even though all checks pass on Netlify. @hellais can you please help check what may be wrong? I wonder if it has to do with the file being named `rus` instead of `ru` (though I think that's unlikely the problem).

sloncocs:
Initially, it was ru (and it is ru in the current PR for Mobile #1285 which has the same issue), I changed it to rus for Desktop in case the URL is sensitive and works only for the right language codes but it didn't help

hellais:
The was merged via: https://github.com/ooni/ooni.org/pull/1288

PR #1287, opened 2022-11-15
Added [Languages.ru] to the config file

sloncocs:


hellais:
The was merged via: https://github.com/ooni/ooni.org/pull/1288

PR #1364, opened 2023-03-22
Add links to the translations to the original files

sloncocs:

### DIFF
--- a/content/about/risks.md
+++ b/content/about/risks.md
@@ -3,8 +3,8 @@
 Translations:
 
 * **Español,** [Riesgos: Lo que debes saber antes de ejecutar OONI Probe](/es/about/risks)
-* **Arabic,** [Risks: Things you should know before running OONI Probe](/ar/about/risks)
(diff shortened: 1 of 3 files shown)


agrabeli (commented):
Perfect, thank you @sloncocs!

My only feedback is that there's some inconsistency, with some languages written in the original language, while others (Arabic and Farsi) in English.

For example:

```
Español, Riesgos: Lo que debes saber antes de ejecutar OONI Probe
Arabic, Risks: Things you should know before running OONI Probe
Farsi, Risks: Things you should know before running OONI Probe
Kiswahili, Hatari: Vitu unavyotakiwa kuvijua kabla ya kutumia OONI Probe
Русский, Риски: что вы должны знать до запуска OONI Probe
```

Would it be possible to have the words `Arabic` and `Farsi`, and the associated document titles, in their respective languages? Or is there an RTL-related challenge?

Otherwise, we could either have all language names in English, or keep it as is for now.

PR #1366, opened 2023-03-30
Outreach kit

hellais:


content/support/ooni-outreach-kit/index.md
```diff
@@ -0,0 +1,104 @@
...
+
+The OONI Outreach Kit includes **materials that you are encouraged to use** as part of your OONI community engagement efforts.
+
+{{<table-of-contents>}}
+
+## OONI Brochures
+
+We provide 2 OONI brochures that include information about using OONI tools and data for investigating internet censorship around the world.
+
+*   **[OONI Brochure (short-version)](<./files/OONI Short Brochure (Interactive\).pdf>)**
```
hellais, line 20: This is an example on how to link the other pages.

Important things to note:
1. Be sure to put the pdf inside of the `files` directory
2. Be sure to wrap the links with `<LINK>` if the filenames have spaces in them
3. If the filename has a `)` character in it, you will have to escape it like this `\)`

PR #1367, opened 2023-04-03
Fixing typos and referencing in iMAP reports 

sloncocs:


bassosimone (commented):
LGTM except for two cases where it seems you don't need to add `^`

sloncocs:
This PR was merged as part of #1368

PR #1372, opened 2023-04-28
Add report on throttling of media during KZ elections

hellais:


agrabeli:
Thank you @hellais and @bassosimone ! I've reviewed and committed some minor md fixes. LGTM. Merging now.

PR #1375, opened 2023-04-28
feat(glossary): define endpoint, dns-over-{https,tls,udp} and udp

bassosimone:
I noticed we needed more terms in the glossary while editing a report we're about to publish.


agrabeli (approved):
Thanks @bassosimone ! LGTM.

PR #1376, opened 2023-04-28
Add brazil telegram report

hellais:


agrabeli:
LGTM

PR #1386, opened 2023-05-19
Add German translation for OONI Probe Mobile user guide

sloncocs:


agrabeli (commented):
Thank you @sloncocs for opening this PR, and huge thanks to those who provided the German translation! Danke schoen! <3

In general, the PR looks good to me. 

My only feedback is:

* If the translators consent, perhaps we can credit them at the start of the translated guide?
* @hellais the Netlify preview doesn't seem to work: https://deploy-preview-1386--ooni.netlify.app/de/support/ooni-probe-mobile Do you perhaps know why? Do you think everything will be OK if we merge?

hellais:
This is superseed by https://github.com/ooni/ooni.org/pull/1395

PR #1394, opened 2023-06-30
Update the Outreach Kit

sloncocs:


agrabeli (approved):
Thanks @sloncocs for updating the Outreach Kit with the new print materials. LGTM! Merging.

PR #1399, opened 2023-07-31
Add report on senegal

hellais:


agrabeli:
Thanks @hellais. The PR looks good to me. I'll merge it (for publication) at 5pm CET today.

agrabeli:
We're holding off from publishing today in light of the latest shutdown order (issued today).

PR #1409, opened 2023-08-22
Add Computech page to the partners list

sloncocs:


agrabeli (approved):
Thanks @sloncocs ! I just committed some edits. Merging now.

PR #1436, opened 2023-11-06
Blog links refresh

akronb:
### Changes in this PR:
- Added a partial that shows the right language title for its shortcode
- Removed unused `archetypes` directory
- Tweaked the way blog links look on the index page – now only the title is a clickable link
- Gave some breathing space between the page title and the links
- Increased the post date font size and made its font color more contrasting
- Increased the font size for translations and revamped the separators between different languages.

Redesign
| before | after |
|--------|--------|
| ooni org_reports_ | 127 0 0 1_1313_reports_ |

### DIFF
--- a/assets/css/master.css
+++ b/assets/css/master.css
@@ -555,31 +555,51 @@ header .wordmark {
 /* ================================================================== */
 /* Blog */
 
-.blog-summary {
+.blog-title {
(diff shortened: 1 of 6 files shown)


akronb:
Also, I found a template for the `/post` page. Do you want to keep it, or can I delete it?

PR #1517, opened 2024-01-12
Add backend dev job opening

hellais:


agrabeli:
Thanks @hellais for integrating my feedback. LGTM!

PR #1549, opened 2024-02-21
Update to the new umami instance

hellais:
This is to be merged and then migration of data should happen


majakomel (commented):
Can we keep `data-domains="ooni.org,ooni.github.io,ooni.netlify.com,openobservatory.github.io"` so previews and similar are not tracked?

hellais:
Ah oops, sorry I missed that part of it. I will re-add it.

majakomel (approved):

PR #1550, opened 2024-02-23
Drop matomo tracking and update privacy policy

hellais:
This PR edits the OONI Data Policy, reflecting that we are replacing the use of [Matomo](https://matomo.org/) with [Umami](https://umami.is/) analytics.


agrabeli (approved):
Thanks @hellais , LGTM!

PR #1570, opened 2024-06-19
Upload TZ Report in Swahili

sloncocs:


agrabeli (approved):
Huge thanks Tori Francis for the Swahili translation of the report! <3

And warm thanks @sloncocs for the coordination and PR! 

I've fixed the ordering of the charts, and everything else looks good. Merging.

PR #1580, opened 2024-07-03
Replace deprecated apt-key with proper keyring in ubuntu/debian installation guide

McLP2:
The deprecation warning annoyed me for a while now, and this is what seems to be the solution.


McLP2:
This will probably fix #1162

PR #1583, opened 2024-07-16
Upload the PDF version of "Technical multi-stakeholder report on Internet shutdowns: The case of Iran amid autumn 2022 protests"

sloncocs:


agrabeli:
Thank you @sloncocs , LGTM!

PR #1591, opened 2024-08-12
Publish Outreach Kit in Farsi

sloncocs:


agrabeli (approved):
Thank you @sloncocs for the PR, and huge thanks to the translators for making the OONI Outreach Kit available in Farsi! <3

PR #1601, opened 2024-08-27
Updated OONI Data Policy for launch of OONI Run v2

agrabeli:
As part of this PR, I have updated the OONI Data Policy to account for changes introduced with OONI Run v2.

We need to **update the date** of the Data Policy based on the launch date of OONI Run v2.

I have already updated the version number of the Data Policy, and that does not need to change.


agrabeli:
We released OONI Probe Android 4.0.0 today with OONI Run v2 support, so I'm merging this PR.

PR #1654, opened 2025-01-22
Update FAQ in English

sloncocs:


agrabeli (approved):
Thanks so much @sloncocs! I just added some edits too. Merging.

PR #1655, opened 2025-01-23
Update dates in FAQ in English

sloncocs:


agrabeli (commented):
Thanks, LGTM.

PR #1657, opened 2025-01-27
Add note about ooniprobe list command on linux

hellais:
closes: https://github.com/ooni/probe/issues/1958


hellais:
This had actually already been done here: https://github.com/ooni/ooni.org/pull/1217/files

PR #1659, opened 2025-01-30
Add issue templates for ooni.org repo

hellais:


hellais:
xref:
* https://github.com/ooni/explorer/pull/985
* https://github.com/ooni/probe/pull/2849
*

PR #1696, opened 2025-02-18
Add post about anonymous credentials

hellais:


mmaker:
Tysm! It looks good, did some minor fixes here and there and pushed everything to 
https://github.com/mmaker/ooni.org/commit/7f77b0a37300902ace894aac2ac0ed828d646259

(I can't seem to be able to push to this branch)
I struggled a bit with `<figcaption>`, which seems to stick with `text-center` from the parent class and can't fix this without touching other parts of your blog.

PR #1698, opened 2025-02-20
Probe security without identification: minor fixes in image displaying and formatting.

mmaker:


hellais:
Thanks! Merging into the based branch

PR #1740, opened 2025-04-02
Update Team page 

sloncocs:


agrabeli:
Thanks @sloncocs for these updates! Could you also add Luis?

sloncocs:
> Thanks @sloncocs for these updates! Could you also add Luis?

done!

PR #1756, opened 2025-04-15
Update OONI Probe mobile user guide (EN)

sloncocs:


content/support/ooni-probe-mobile/images/image118.jpg
aanorbel, line 1: We should probably redo this screenshot

content/support/ooni-probe-mobile/images/image51.jpg
aanorbel, line 1: I dont think the first element in the list, `Websites` should be highlighted

content/support/ooni-probe-mobile/images/image154.jpg
aanorbel, line 1: I am not sure performance failing in this screenshot is intentional. 
I can provide the same screenshot we used in the store so its more consitent.

aanorbel (commented):
The guide looks amazing so far. 

My recommendation is to have a somewhat consistent size for the mobile screenshots (the width).

content/support/ooni-probe-mobile/images/image181.jpg
sdsantos, line 1: The copy is now different on this screen. We need to retake the screenshot.

sdsantos (commented):
The index above works for sending the user to a specific section. But it would be nice if the section headers would also be linked to the anchor, so we could just click it and copy the URL.

`### [Manually measuring the blocking of instant messaging apps](#manually-measuring-the-blocking-of-instant-messaging-apps)`

But maybe the markdown processor we're using supports:

`### [Manually measuring the blocking of instant messaging apps][]`

---

> Step 2. Depending on your operating system (Android, iOS, F-Droid), tap the relevant app store button.

Google Play/F-Droid apply to the same operative system, Android. The choice in that case it's more of convenience vs privacy. Not a big deal, since most people that know F-Droid will know if they want to use that option. But the phrase is not entirely correct, so I wanted to raise that.

---

> Step 6. If you would like the app to collect the measurements automatically, tap the Sounds great button.

The button now only says "Yes"

---

> Step 9. Tap OK.

The button now says "Let's go".

---

> Installed OONI Run links. Beneath all the cards you will see a section with the OONI Run links you have installed on your app. You can read more about OONI Run links here.

Note: You will only see this is there's any installed link. After onboarding, the user won't have any.

---

The section "Manually measuring the blocking of circumvention tools" seems to be missing the "You are now running OONI’s..." paragraph.

---

> In the above example, we have enabled 5 categories (“LGBTQI”, “Political Criticism”, “News Media”, “Human Rights”, “Communication tools”).

The screenshot has:
- LGBTQ+ (slightly different acronym)
- Human Rights Issues (Issues at the end)
- No "Communication tools"

---

The "Install the link" screen looks a bit different now, but instructions are still correct.

---

For "Deleting test results" you can also use the clear button on the top right corner of the "Test Results" screen. Clearing from the Settings clears more stuff like logs. But in practice, the user won't tell the difference.

---

The Settings "Notifications" screen has disappeared with https://github.com/ooni/probe-multiplatform/pull/652

content/support/ooni-probe-mobile/images/image154.jpg
sloncocs: I don't think it is a problem that it is failing because users may see this result as well, and if we update this screenshot, we will need to update all screenshots in this section

content/support/ooni-probe-mobile/images/image181.jpg
sloncocs: done!

sloncocs:
I addressed all content-related feedback. 

@majakomel would you have a moment to look into markdown and image formatting? Maybe there is a way to adjust the mobile screenshots to the same size automatically?

PR #1844, opened 2025-08-30
Publish GG OMG Village announcement

sloncocs:


agrabeli (approved):
@sloncocs thanks again for opening the PR! I made various commits with md improvements and trying to fix the tables (one of which I broke). I think the PR is now ready for being merged.

PR #1870, opened 2025-12-08
Replaced twitter link to x link

Sinhapratik006:
Replaced twitter old link to new X link


layouts/post/single.html
majakomel: It seems no actual changes have been added to this file, can you remove it from PR?

PR #1909, opened 2026-06-30
Research blog post on assessing the effectiveness of anonymous credentials

LDiazN:


agrabeli (commented):
Thank you @LDiazN for preparing this PR, it looks great to me! 

My only feedback is the following:
* Update the date to include the true date of publication (it currently says 26th June 2026)
* Some of the charts say "Figure 3" and "Figure 5", and I couldn't find Figure 4, etc. This could easily be addressed by replacing "Figure X" with generic "Chart". It might also be worth being consistent in including chart descriptions for other figures too. 

Thanks again!

PR #1914, opened 2026-07-21
Remove use of run widget

majakomel:


agrabeli:
Thanks so much @majakomel for preparing this PR! Just reviewed it, and it looks good to me. I only made some minor commits to remove some sentences referencing the OONI Run widget (and linking to an old Security Without Borders post which no longer exists). I think this PR is ready for merging.

